PR #50, opened 2014-02-20
Warn on fid light in acq search box

jeanconn:
Warn on fid light in acq search box to avoid obsid 14577 issue.  This should be a "yellow" warning for an acq-only star, red for a acq/guide star.

### DIFF
--- a/src/lib/Ska/Starcheck/Obsid.pm
+++ b/src/lib/Ska/Starcheck/Obsid.pm
@@ -1140,7 +1140,7 @@ sub check_star_catalog {
 
         # Check for situation that occurred for obsid 14577 with a fid light
         # inside the search box (PR #50).
-        if ($type =~ /BOT|GUI|ACQ/){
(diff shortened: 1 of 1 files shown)


taldcroft:
> Could also be a comment that refers back to the PR. Imagine me in 3 years looking at a test and trying to understand what's going on.

This just happened, and PR #50 was quite useful!!

taldcroft:
> With all these corner cases and special rules, it is not a bad idea to have some details in the code comments.

And detailed code comment with the provenance of the change was useful as well.

jeanconn:
Hah.  If you are in fids we've also got issue #243 at some point (so fids need a bit more review in warnings and severity for 13.1).

taldcroft:
BTW, I don't remember why this is yellow (not orange) or an ACQ-only star.

jeanconn:
Um.. That *is* issue #243 (aka Jean suggests that ACQ-only should be orange not yellow).

PR #77, opened 2014-06-12
Plot catalogs and fields with matplotlib

jeanconn:
Plot starcheck catalogs in matplotlib.


Makefile
```diff
@@ -23,8 +23,9 @@ TEST_BACKSTOP = AUG0104A/CR214_0300.backstop
 DATA_FILES = starcheck_data/aca_spec.json starcheck_data/ACABadPixels starcheck_data/agasc.bad \
 	starcheck_data/fid_CHARACTERIS_JUL01 starcheck_data/fid_CHARACTERIS_FEB07 \
 	starcheck_data/fid_CHARACTERISTICS starcheck_data/characteristics.yaml \
-	starcheck_data/A.tlr starcheck_data/B.tlr starcheck_data/tlr.cfg
-
+	starcheck_data/A.tlr starcheck_data/B.tlr starcheck_data/tlr.cfg \
+	starcheck_data/overlib.js starcheck_data/up.gif starcheck_data/down.gif \
+	starcheck_data/D3_LICENSE starcheck_data/d3.js starcheck_data/d3.min.js
```
jeanconn: I should cut the D3 stuff as I'm planning this as a matplotlib/png PR with a follow-up sometime in the future for links to interactive plots.

starcheck/plot.py
```diff
@@ -0,0 +1,187 @@
...
+plt.rcParams['savefig.facecolor'] = backcolor
+plt.rcParams['savefig.edgecolor'] = backcolor
+
+
+def symsize(mag):
+    # map mags to figsizes, defining
+    # mag 6 as 20 and mag 11 as 1
+    # interp should leave it at the bounding value outside
+    # the range
+    return np.interp(mag, [6.0, 11.0], [20.0, 1])
```
jeanconn: These don't scale appropriately for the "field" view

star_view_60110

jeanconn:
@taldcroft  what plot edits need to happen to the PNG output to make this acceptable for flight use?

stars_5353

starcheck/plot.py
jeanconn: Might be simplest to revert the change I added for a different figure sizes and just image resize the field plot.

starcheck/plot.py
```diff
@@ -0,0 +1,187 @@
+import os
```
jeanconn: This module is largely copied in from mica, but will probably end up living in starcheck when complete.

src/starcheck.pl
```diff
@@ -473,34 +482,55 @@
...
+                             starcat_time=>"$obs{$obsid}->{date}",
+                             outdir=>$STARCHECK);
+            plot_cat_wrapper(\%plot_args);
+            alarm 0;
+        };
+        if ($@){
+            push @global_warn, "ERROR: $@";
+            die "$@\n";
+        }
+        alarm 0;
```
taldcroft: What about having one initial alarm'ed check of loading Python, with various required libraries being imported.  After that it would just call the routines and wait as long as needed.  Basically this is non-optional so it has to work.  If the Perl - Python interface is unreliable so that you can't know if a Python routine ran successfully to completion (and matplotlib will always create the plot if you), then I don't know what to say.

starcheck/plot.py
```diff
@@ -0,0 +1,187 @@
...
+        for star in bright_field:
+            yag, zag = radec2yagzag(star['RA_PMCORR'],
+                                    star['DEC_PMCORR'],
+                                    quat)
+            yag *= 3600
+            zag *= 3600
+            yags.append(yag)
+            zags.append(zag)
+
+        bright_field = Ska.Numpy.add_column(bright_field,
```
taldcroft: Use astropy Table here (convert field at the top).  All the Ska.Numpy structured array manipulation functions are deprecated.

starcheck/plot.py
```diff
@@ -0,0 +1,187 @@
...
+
+def symsize(mag):
+    # map mags to figsizes, defining
+    # mag 6 as 20 and mag 11 as 1
+    # interp should leave it at the bounding value outside
+    # the range
+    return np.interp(mag, [6.0, 11.0], [20.0, 1])
+
+
+def star_plot(catalog=None, quat=None, field=None, title=None, figsize=(5, 5)):
```
taldcroft: Doc string.

Variable names like `quat

src/starcheck.pl
jeanconn: I was planning to get rid of the alarms altogether.  I don't see what value they add at this point.  And regarding the Perl checks to see if files are there, you are right and that is both old and an old comment... I don't see much need to check to see if the plots are there if there are no errors  (if I find ways to break Inline::Python in testing where the plotting dies silently, I'll rethink this).

jeanconn:
Regarding dynamic plots, zooms with the inverted axis still seem broken in D3/mpld3.  If we'd like dynamic plots, I don't know if we'll want to look again at other libraries (bokeh) or do something like invert the yang data values and plot with a dummy axis or similar hack.

starcheck/plot.py
jeanconn: Yes, this all needs rework now that it is more than a year later.  But is the plot itself kinda what we want?  I think I can work on the rest without detailed code review just yet.

starcheck/plot.py
```diff
@@ -0,0 +1,187 @@
...
+
+        bright_field = Ska.Numpy.add_column(bright_field,
+                                            'yang',
+                                            yags)
+        bright_field = Ska.Numpy.add_column(bright_field,
+                                            'zang',
+                                            zags)
+        color = np.ones(len(bright_field), dtype='|S10')
+        color[:] = 'red'
+        faint = ((bright_field['CLASS'] == 0)
```
taldcroft: The faint limit should be a function kwarg with default 10.7.  I can imagine setting this faint limit dynamically to be the red warning limit or maybe red warning + 0.3 or something.

starcheck/plot.py
```diff
@@ -0,0 +1,187 @@
...
+        ax.scatter(fid['yang'], fid['zang'],
+                   facecolors=face,
+                   edgecolors='green',
+                   marker='+',
+                   linewidth=3,
+                   s=200)
+
+    # plot field if present
+    if field is not None:
+        faint_plot_mag = 11.0
```
taldcroft: In general I would prefer using a comment line over a one-off variable:
```
# Filter out faint stars
bright_field = field[field['MAG_ACA'] < 11.0]
```
But in fact one of two things should be changed anyway.  Either:

- Allow for `faint_plot_mag` to be a function kwarg  OR
- Assume that the caller has filtered appropriately.

I would go for the function kwarg.

Finally, notice that in the rest of the function you never use `field` again.  When you are filtering a table in a setting like this and don't need the original, then just filter in place.  So I would have:
```
field_stars = field_stars[field_stars['MAG_ACA'] < faint_plot_mag]
```
Someday you'll be able to write:
```
field_stars = field_stars.where(f'MAG_ACA < {faint_plot_mag}')
```

jeanconn:
Still need to implement compass plot.

starcheck/plot.py
```diff
@@ -0,0 +1,187 @@
...
+        bright_field = field[field['MAG_ACA'] < faint_plot_mag]
+        yags = []
+        zags = []
+        for star in bright_field:
+            yag, zag = radec2yagzag(star['RA_PMCORR'],
```
taldcroft: I won't say you can't do this, but the always-one-arg per line style doesn't always contribute to readability for me, esp with just a few short args.

taldcroft: Always search for ways to use functional forms and tables.
```
# Add star Y angle and Z angle in arcsec to the field_stars table
yagzags = (radec2yagzag(star['RA_PMCORR'], star['DEC_PMCORR'], attitude) 
           for star in field_stars)
yagzags = Table(rows=[(y * 3600, z * 3600) for y, z in yagzags], names=['yag', 'zag'])
field_stars = table.hstack([field_stars, yagzags])
```
So 4 lines with a comment that clearly states what is being done.

starcheck/plot.py
```diff
@@ -0,0 +1,187 @@
...
+            yags.append(yag)
+            zags.append(zag)
+
+        bright_field = Ska.Numpy.add_column(bright_field,
+                                            'yang',
+                                            yags)
+        bright_field = Ska.Numpy.add_column(bright_field,
+                                            'zang',
+                                            zags)
+        color = np.ones(len(bright_field), dtype='|S10')
```
taldcroft: Maybe you can use an object `'O'` type here and not have to pre-declare the longest possible color name.

starcheck/plot.py
jeanconn: Should it be a kwarg or should I make field an object that has that property or some such?  Not sure which might make more sense for interface.

taldcroft: I don't think this needs an object class.

taldcroft:
On the original question of plot edits required, I think that the original color scheme, line thicknesses and relative sizing of plot elements are good.  So work to replicate that EXCEPT that the change from magenta to orange is a good one and should stay.  The title should be the same except without the SAUSAGE disclaimer.  Also the readout register (double lines on the sides) needs to be replicated.

taldcroft:
It might be worth adding a light line (maybe in the magenta-ish range) showing the quadrant boundaries.

jeanconn:
I think I have the relative star symbol sizes at this point, agreed?  Or more work required there.? 

That leaves: refine line thicknesses, match color scheme, fix guide circle and fid symbol sizes.

I had the readout register at native display in matplotlib, but looks like that will need tweaks to show up in the png. 

Thanks!

jeanconn:
The quadrant boundaries really aren't lines in yag and zag though...  Light magenta connect-the-dots?

taldcroft:
Yes, connect-the-dots.  This more or less applies to all the lines since everything is fundamentally square only in pixels.  At the resolution of a single PNG plot for starcheck, none of this matters.  It's only if zooming were allowed that there would be an issue.

The mpld3 issue is annoying.  This plot is fairly simple so it might be straightforward to do with bokeh.  Might be worth giving that a shot.

taldcroft:
BTW, even though this lives in starcheck it will be definitely be useful in several other contexts.  Keep that in mind in the design.  So far what I've seen is consistent with generic use as a utility.

jeanconn:
Also, if we're not plotting the faint stars, the magenta vs orange won't matter much for starcheck output (or did you want a default of 10.7 for the routine but for starcheck to pass a different limit).

jeanconn:
@taldcroft  am I getting warmer after this last set of updates?  Grid or no grid on the plots?  What's the appropriate faint limit?

stars_5353

(flight output for reference):

stars_5353

taldcroft:
In theory the faint limit should be related to the star mag uncertainty since the main driver for showing them is as spoilers.  Perhaps there should be a `red_mag_limit` input parameter and every (otherwise acceptable) star bright than `red_mag_limit` is plotted in black.  Then any star with:
```
mag_delta = max(1.5, 3 * mag_err)  # max of 1.5 mag or 3-sigma
possible_spoiler = (mag > red_mag_limit) and (mag < red_mag_limit + mag_delta)
if possible_spoiler:
   # plot in orange as a possible spoiler star
```
I'm not sure about the exact value for the 1.5.  There is probably something like this in star selection.

taldcroft:
The plot looks pretty good now.  The 6.0 mag filled circle is a fair bit smaller though, so that should be fixed to match more closely.  

I like the grid.  

One enhancement might be to play around with the alpha, maybe something like 0.7 or so to make it easier to see overplots.  This might not work out, it's just a thought.

taldcroft:
Now that I look closer it seems like you might already be using an alpha for the text and fids.  It's definitely fainter than in the perl version (but it looks OK).

taldcroft:
Maybe we should remove the "Stars at" in the title, since it is a bit long and the plot is not just stars.

jeanconn:
Regarding the text and fids, I didn't use an alpha, but I changed the annotation "weight" to light and the linewidths in the markers to .5.  I'm not sure exactly how each effect gets rendered, but it ends up looking a bit like an alpha.

jeanconn:
Regarding the faint limit and mag error, I think we may want to update the miniagasc to just have all the stars in the flight agasc.  Basically, I think that your filter on

idxs = tbl.getWhereList("(MAG_ACA - (3.0 * MAG_ACA_ERR / 100.0)) < 11.5")

was not exactly what SAUSAGE needed for spoiler rejection, and we might also want some of those stars around for this.  I'll double-check this.

jeanconn:
Todo:

 - [ ] Refine/define faint limits

Todo outside PR:

 - [ ] Remove alarms/timeouts

taldcroft:
What does SAUSAGE need?  I think this probably corresponds to the stage 0 rule, which I thought was 3-sigma (but maybe 3-sigma for both stars?).  My guess is that a high fraction of the time the current miniagasc sampling is sufficient, but maybe that guess is wrong.

jeanconn:
I think I correctly applied the rule here:

https://github.com/sot/aca_required_temp/blob/dc0eb3ebcf3aa41dfdd3beb5a486a1096eef0501/mini_sausage.py#L124

which is using 3 times the sqrt of the sum of the square of the individual star errors.  Also, magdifflim is set at -inf in the characteristics, so no star can be too dim to be considered for the slope/intercept spoiler match. 

And I'm sure you are right that a high fraction of the time the current sampling is basically sufficient, but in the 2 obsid examples I thoroughly debugged in SAUSAGE it was an issue and I wanted exact matches.  I should actually be able to do some comparisons in aggregate to see how much of an issue it is (but since it was easy enough to just use all the stars, I went in that direction).

jeanconn:
Getting back to the plotted faint limit, why would we call this `red_mag_limit`?  I'm a little confused about the plotted black limit, orange limit, and red limit.

taldcroft:
- Black is meets all criteria for an acq star, including brighter than `red_mag_limit` (the 50% probability-of-acquisition magnitude).
- Orange meets all criteria for an acq star except it is fainter than `red_mag_limit` (and brighter than the limit discussed earlier for a spoiler star).  These are stars one might choose if absolutely nothing else is available, but will have < 50% chance of success.
- Magenta fails at least one criteria for an acq star (without considering magnitude).

jeanconn:
For "meets all criteria for an acq star", I don't have all of the acq star
checks in Python, but I assume the subset of checks solely based on AGASC
properties (excluding position checks) will do?

taldcroft:
Yes you know what I meant.

jeanconn:
Was just concerned that the colors would look more like SAUSAGE (minus the
green "next best" candidates) and didn't know if you really might want to
go all the way down that path yet.

taldcroft:
Just to be clear, the mag limit for magenta is the same formula as for orange.  Final colors TDB based on how it looks.  Note that for starcheck catalog evaluation we'll have the upper-right black & white plot showing all stars in the field.  (Open question - use the same color scheme but with no upper mag limit?  Probably worth trying just for kicks).

jeanconn:
I'm over-complicating this I'm sure, but I'm still not sure how to determine this limit:

"(and brighter than the limit discussed earlier for a spoiler star)"

if the SAUSAGE method was to use the 3 sigma * np.sqrt(stderr2[cand_idx] + stderr2[cat_idx])).

jeanconn:
stars_4592
star_view_4592

starcheck/plot.py
```diff
@@ -0,0 +1,327 @@
...
+    # get the agasc field once and then use it for both plots that have stars
+    field_stars = agasc.get_agasc_cone(ra, dec,
+                                       radius=1.5,
+                                       date=DateTime(starcat_time).date)
+
+    # Mark bad stars from current limits in characteristics
+    bad_stars = ((field_stars['CLASS'] != 0) | (field_stars['MAG_ACA_ERR'] > 100) |
+                 (field_stars['POS_ERR'] > 3000) | (field_stars['ASPQ1'] > 0) |
+                 (field_stars['ASPQ2'] > 0) | (field_stars['ASPQ3'] > 999) |
+                 (field_stars['VAR'] > -9999))
```
jeanconn: If we move the plotting code to chandra_aca, do you think this particular set of bad_star criteria should go there too as a default filter?  Or should this stay in starcheck?  My thought was that this "make_plots_for_obsid" routine would stay in starcheck.

jeanconn:
stars_4592

taldcroft:
Maybe bring up the background stars just a touch.

jeanconn:
stars_4592
star_view_4592

taldcroft:
It looks like there are orange and small brown-ish stars.  I don't know what brown-ish is unless it is orange overplotted once.

jeanconn:
Does that mean you don't like the colors,  don't understand their
combinations,  or don't like the rendering?  I was seeing two tomato stars
looking a little more purple than one tomato star and a tomato and
lightseagreen combination looking a little browner.  But I haven't looked
at intentionally combining large blobs to understand the full range of
possible colors.

taldcroft:
I just thought there were going to be only 3 colors (OK=black, above red_mag but within 3 sigma=cyan, bad=red(ish)).  Did I get confused?  I have to run otherwise I would just look at the code again.

jeanconn:
There are only three starting colors (with the rough color names you
mention),  but with the alpha and some tiny dot rendering color oddities we
do get a bit of a range.  I'll plot the range.  We could also just lighten
the colors a bit and chuck the alpha.

taldcroft:
I was mostly wondering if there might be  a bug and something getting double plotted. I still want to see the main start color logic rewritten so I can understand it.

jeanconn:
Sure.  And you are probably right that dim and bad are getting over
plotted.   I'm not sure if that is a bug, but if it is logically not what
you expect then the code will reflect that too.

taldcroft:
:+1:

Issue #97, opened 2015-03-03
Fix HTML format around RDX obs so Next/Prev links don't move left/right


jeanconn:

taldcroft:
:+1: :+1: :+1: :+1: :+1:

jeanconn:
Closed by #104

PR #104, opened 2015-04-03
Fix formatting

jeanconn:
Fix some fixed-width formatting hiccups in starcheck output.  These fixes should stop the Prev/Next link/buttons from moving around.


jeanconn:
Closes #97

Issue #15, opened 2012-07-13
Remove benign dark current transponder warnings


jeanconn:
Warnings like:

Transponder state change during replica at  2012:198:01:14:53.985
in 

https://icxc.harvard.edu/mp/mplogs/2012/JUL1412/oflsa/starcheck/dark_cal_super_verbose.html

Since the code has already established the desired transponder state, commands that are consistent with that state should be ignored instead of producing warnings about a change of state.  In this case, the replica is planned for transponder B / CIU512X.  A CIU512X CIMODESL before the replica does not exist in the loads and is expected in real-time before replica 0.  While the code should still warn that the setup requires real-time commanding, it should not warn about the expected/fine CIU512X CIMODESL at the end of the replica.  Warnings should only be generated when transponder commands exist within the replica that are inconsistent with the desired state.

@taldcroft

jeanconn:
OBE

PR #105, opened 2015-04-03
Reduce benign dark cal transponder warnings

jeanconn:
The code was set up to warn on any transponder commanding that changed
the apparent transponder state during a replica.  This included, say,
CIMODESL commands that were consistent with the selected replica, but
not consistent with the load-commanded state before the replica.  The new
code tracks "state" changes as well as the states and only throws warnings
on new commands if they are actually inconsistent with the transponder that
is expected for the replica.


jeanconn:
Closes #15

jeanconn:
Seems OBE

PR #109, opened 2015-05-15
Remove yaml

jeanconn:
Remove YAML config file and use of YAML in Python and Perl code.


taldcroft:
What's the driver for this?  I guess we soured on YAML a bit ago but I've been influenced by others to regard the human-readability of config files as a big plus.  JSON is kinda finicky to edit and less easy to read.

jeanconn:
I profiled calc_ccd_temps and yaml calls (even trying the libyaml stuff)
seemed to be 3/4 of the CPU time.

We already have the JSON dependence and the characteristics file is really
small.

jeanconn:
And when starcheck is all Python, I'd be inclined to write characteristics
as python code instead if YAML.

jeanconn:
And by 'seemed to be 3/4 of the CPU time ' I mean that removing YAML
brought the code down to 12s from 40s in my MAY1815 test case.

Obviously not the 10 minute network culprit.

taldcroft:
There has to be something else happening.  It simply can't be taking 28s to parse some small YAML files.

jeanconn:
OK, if you don't want to convert the characteristics file, I'm fine with that.  I just figured any import we removed was likely to speed things up.

The fundamental speed increase here I suppose came not from stripping out yaml entirely but from moving back to actually using the json loader to load the *json* file.  
Previously, the json file of obsids was read by the yaml loader, as I thought the subset was all we needed and it didn't return yucky looking unicode... but I'd guess the time is lost in casting objects?  I can look into this some more if you like.

```
$ cat pyyaml.py
from yaml import load
obsids = load(open("obsids.json").read())
$ time python pyyaml.py 
real	0m7.709s
user	0m7.589s
sys	0m0.102s

$ cat libyaml.py 
from yaml import load
from yaml import CLoader as Loader
obsids = load(open("obsids.json").read())
$ time python libyaml.py 
real	0m7.691s
user	0m7.558s
sys	0m0.112s

$ cat impjson.py
import json
obsids = json.loads(open("obsids.json").read())
$ time python impjson.py
real	0m0.052s
user	0m0.039s
sys	0m0.009s

**Comment by taldcroft** [comment]
Interesting. One thing is that you need Loader=Loader in the Cloader version call.

**Comment by jeanconn** [comment]
Ah,  I figured when I didn't see any improvement that wasn't working...  With Loader=Loader:

```
(test)bash-4.1$ time python libyaml.py 
real	0m0.972s
user	0m0.926s
sys	0m0.039s
```

taldcroft:
http://stackoverflow.com/questions/2451732/how-is-it-that-json-serialization-is-so-much-faster-than-yaml-serialization-in-p

taldcroft:
But regardless you are right that the JSON obsids file should be read by the JSON parser in python.

jeanconn:
And then leave the (hardly-used) characteristics file as YAML or not then?

taldcroft:
Yeah, just leave the config file as YAML.

jeanconn:
Superceded by #111 which leaves the yaml config file alone.

PR #111, opened 2015-05-28
Load json with json instead of yaml parser

jeanconn:
This seems faster than yaml even if libyaml is used.

### DIFF
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -133,9 +133,11 @@ def get_ccd_temps(oflsdir, outdir='out',
     # save spec file in out directory
     shutil.copy(model_spec, outdir)
 
-    # the yaml subset of json is sufficient for the
-    # JSON we have from the Perl code
-    sc_obsids = yaml.load(json_obsids)
+    # json_obsids can be either a string or a file-like object.  Try those options in order.
+    try:
+        sc_obsids = json.loads(json_obsids)
+    except TypeError:
+        sc_obsids = json.load(json_obsids)
 
     tnow = DateTime().secs
 


taldcroft:
Looks fine, but the code comment almost increases my confusion.  It sounds like you are editorializing about the json API, and saying that the unicode string is from stdin got me confused since the function arg default is sys.stdin.  So try to be succinct and direct in the comment:
```
# json_obsids can be either a string or a file-like object.  Try those options in order.
```

jeanconn:
I suppose I was editorializing about the API, as yaml.load handled both fine, but OK.

jeanconn:
Though I don't entirely understand "saying that the unicode string is from stdin got me confused since the function arg default is sys.stdin."  Which vocabulary was confusing in there?

jeanconn:
The comments were also supposed to help me remember that loads() takes a str and load() takes a file-like object without having to go and reference the JSON help again.  I thought that was a trick about the API that was worth noting for myself.

taldcroft:
What was confusing is that the function `get_ccd_temps` can take `json_obsids` as a string (which doesn't have to be unicode) or a file-like object, and it is irrelevant where the string originally came from.  But then `get_opt` only allows `json_obsids` as a file-like object with the default being sys.stdin.  So when I first read the comment, knowing about the main `get_opt` only allowing file-like with a default of stdin, and the misdirection in the comment about the string coming from stdin, it actually took me more time to grok the full situation.

jeanconn:
Ah.  I think I understand now. Regarding "Which vocabulary was confusing in there?" the problem is that I was referring to the string passing direct from starcheck.pl as a stdio operation, which was incorrect.

And while it "is irrelevant where the string originally came from" it is helpful to know which code paths are called by which methods of operation to know what to test.  In this case, calc_ccd_temps has support for stdin, a file, and a string, and all three should be tested... and maybe someday there will be tests...

taldcroft:
:+1: to merge

PR #103, opened 2015-05-28
Tighten star mag limits when CCD temps warmer than -14

jeanconn:
Tom proposes:

decrease mag limits by 0.15 for each 1 C increase above -14 C

### DIFF
--- a/src/lib/Ska/Starcheck/Obsid.pm
+++ b/src/lib/Ska/Starcheck/Obsid.pm
@@ -935,6 +935,17 @@ sub check_star_catalog {
 	$dither = 20.0;
     }
 
+    # Adjust mag limits based on CCD temperature
+    if ($self->{ccd_temp} > -14){
+        my $mag_delta = ($self->{ccd_temp} - -14) * 0.15;
+        $mag_faint_red -= $mag_delta;
+        $mag_faint_yellow -= $mag_delta;
+    }
+    # Save these limit values
+    $self->{mag_faint_yellow} = $mag_faint_yellow;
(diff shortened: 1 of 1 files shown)


taldcroft:
As far as code goes this looks fine.  I think it would be worth discussing this with SSAWG before actually merging though.

Also, I was wondering if it might be enlightening to run the regression tests for recent years *without* the additional dynamic mag limits line in the output in order to get a cleaner idea of the diffs in yellow and red star warnings.

taldcroft:
I should note that I haven't explicitly verified that the current set of loops in front of the CCD temp. calculation is sufficient for populating the obs structure.  I'm assuming that if you got a good run then that is the case.

jeanconn:
Regarding waiting for SSAWG, are you concerned with the determination of the "scale factor" or the whole idea?

jeanconn:
And yes, I did functionally test this before make a PR.

taldcroft:
Mostly the scale factor and whether we should put in some conservatism.  Of course it has also occurred to me that this approach is quite simplistic and not robust against accumulating radiation damage.  The acquisition probabilities model could help. However, getting something simple in place sooner rather than later is a good thing.

jeanconn:
Maybe I should just do the main "loop" reorganization as a separate PR.  It add the option to use the temperature values in the per-obsid checking, but the implementation of the actual mag scaling should probably be separated out anyway.

taldcroft:
Good idea.

jeanconn:
Reorganization now in #113

jeanconn:
Superceded by work in #119

PR #113, opened 2015-06-08
Calculate temperatures before doing main starcheck checking.

jeanconn:
Rearrange loops to determine ccd temperatures via the model earlier in the overall starcheck logic.  This makes the temperatures available to the main starcheck catalog checking as needed.


taldcroft:
:+1: assuming you do some subset of regression tests and see no diffs.

jeanconn:
Full regression test just shows change in order of warnings related to the change in the actual order of those checks in the code.

Issue #85, opened 2014-08-18
Prob of N or fewer stars is actually prob of fewer than n stars


jeanconn:
This line 

https://github.com/sot/starcheck/blob/09dd3cdff6dcd3b3900def425c04ff9cac7b937c/src/lib/Ska/Starcheck/FigureOfMerit.pm#L137

should probably loop from $j = $i + 1 to $n_prob if we want N or fewer from the next loop (which does log(1 - the value of this loop))

jeanconn:
Fixed in #117

PR #117, opened 2015-07-08
Functions related to star acquisition and guide tracking probabilities

taldcroft:


starcheck/star_probs.py
```diff
@@ -0,0 +1,221 @@
...
+
+    :param date: Date(s) (scalar or np.ndarray, default=NOW)
+    :param t_ccd: CD temperature(s) (degC, scalar or np.ndarray, default=-19C)
+    :param mag: Star magnitude(s) (scalar or np.ndarray, default=10.0)
+    :param color: Star color(s) (scalar or np.ndarray, default=0.6)
+    :param spoil: Star spoiled (boolean or np.ndarray, default=False)
+
+    :returns: Acquisition success probability(s)
+    """
+    from mica.archive.aca_dark.dark_model import get_warm_fracs
```
taldcroft: This is a perfect example of why localized imports are dangerous, and how I need to follow my own advice!  Before the star working group I had scanned the imports for this file (at the top) and had made a mental note that there were no real dependencies that don't have Matlab equivalents.  WRONG.

starcheck/star_probs.py
```diff
@@ -0,0 +1,221 @@
...
+SOTA_FIT_ONLY_1P5 = [8.541709287866361,
+                     0.44482688155644085,
+                     -3.5137852251178465,
+                     -1.3505424393223699,
+                     1.5278061271148755,
+                     0.30973569068842272]
+
+
+def t_ccd_warm_limit(date, mags, colors=0,
+                     min_n_acq=5.0,
```
jeanconn: Do you have a plan to update this to use a probability value instead of min_n_acq to determine warm limit?  I'd like to use a version with that in the LTS checker soon if that is the way we want to go in starcheck.

taldcroft: After I get the aimpoint stuff in order this will be next.

jeanconn: And one of the reasons I hadn't released a standalone tool for SOT MP is that it my LTS evaluator code can only run in a dev ska until some of these functions are finalized and installed someplace reasonable.

taldcroft: This is now a very high priority for me.

src/lib/Ska/Starcheck/FigureOfMerit.pm
```diff
@@ -25,45 +21,74 @@ require Exporter;
...
-
-    #calculate the probability of acquiring n stars
-    my $n_slot = $#probs+1;
-    my @nacq_probs = prob_nstars($n_slot, @probs);
-    #calculate the probability of acquiring at least n stars
-    $self->{figure_of_merit} = cum_prob($n_slot, @nacq_probs);
     $self->{acq_probs} = \%slot_probs;
 
+    # Calculate the probability of acquiring n stars
+    my ($n_acq_probs, $n_or_fewer_probs) = _prob_n_acq(\@probs);
```
jeanconn: Does this also resolve #85 ?

taldcroft: Yes.  N_or_fewer is exactly that.

jeanconn: Would it be valuable also then to include the probability of 1 or fewer in the starcheck block of text?  I think we'd talked about that in the probability SSAWG meeting as the easiest way to line up the probabilities from the tool directly with the estimated probability of a safing action based on acquisition failure.

src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1660,15 +1659,15 @@ sub print_report {
     if ($c = find_command($self, "MP_STARCAT")) {
 
 
-	my @fid_fields = qw (TYPE  SIZE MINMAG GS_MAG MAXMAG YANG ZANG DIMDTS RESTRK HALFW GS_PASS GS_NOTES);
+	my @fid_fields = qw (TYPE  SIZE P_SUCC GS_MAG MAXMAG YANG ZANG DIMDTS RESTRK HALFW GS_PASS GS_NOTES);
 	my @fid_format = ( '%6s',   '%5s',  '%8.3f',    '%8s',  '%8.3f',  '%7d',  '%7d',    '%4d',    '%4d',   '%5d',     '%6s',  '%4s');
-	my @star_fields = qw (   TYPE  SIZE MINMAG GS_MAG MAXMAG YANG ZANG DIMDTS RESTRK HALFW GS_PASS GS_NOTES);
+	my @star_fields = qw (   TYPE  SIZE P_SUCC GS_MAG MAXMAG YANG ZANG DIMDTS RESTRK HALFW GS_PASS GS_NOTES);
```
jeanconn: Change P_SUCC to P_ACQ

jeanconn:
Todo:

  - [ ] Color code the acquisition probability and the mag in the catalog table based on the acquisition probability.
  - [ ] Print '---' in acquisition probability for FID/GUI/MON
  - [ ] Set threshold for red probability printing at min_n_acq=(2, 8e-3)

jeanconn:
@taldcroft  For the color coding in the catalog table, for acq / bot stars I'll use the 50%, 75% acquisition probabilities.  Should I just use the Red/Yellow mag limits to determine colorization for guide stars?

taldcroft:
Sure.

src/lib/Ska/Starcheck/FigureOfMerit.pm
jeanconn: Per SSAWG discussion, leave this at 2,3, or 4.

src/lib/Ska/Starcheck/FigureOfMerit.pm
```diff
@@ -25,45 +21,78 @@ require Exporter;
...
+def _acq_success_prob(date, t_ccd, mag, color, spoiler):
+    out = acq_success_prob(date, float(t_ccd), float(mag), float(color), spoiler)
+    return out.tolist()
 
+def _prob_n_acq(acq_probs):
+    n_acq_probs, n_or_fewer_probs = prob_n_acq(acq_probs)
+    return n_acq_probs.tolist(), n_or_fewer_probs.tolist()
+};
+
+our $CUM_PROB_LIMIT = 8e-3;
```
jeanconn: @taldcroft Is this what you had in mind for the limit?

src/lib/Ska/Starcheck/FigureOfMerit.pm
```diff
@@ -25,45 +21,78 @@ require Exporter;
...
 
+    # Calculate the probability of acquiring n stars
+    my ($n_acq_probs, $n_or_fewer_probs) = _prob_n_acq(\@probs);
+
+    $self->{figure_of_merit} = {expected => substr(sum(@probs), 0, 4),
+                                cum_prob => [map { log($_) / log(10.0) } @{$n_or_fewer_probs}],
+                                cum_prob_bad => ($n_or_fewer_probs->[2] > $CUM_PROB_LIMIT)
+                                };
+    if ($n_or_fewer_probs->[2] > $CUM_PROB_LIMIT){
+        push @{$self->{warn}}, "WARNING: Probability of 2 or fewer stars > $CUM_PROB_LIMIT\n";
```
jeanconn: And do you mind if I just add a warning to the block if we exceed the limit?  This should help it show up as something of concern in the top level summary.

src/lib/Ska/Starcheck/FigureOfMerit.pm
jeanconn: I haven't been able to find any catalogs that exceed it (as in A products) to have a good before/after test.

taldcroft: Based on this image from ref below, 8e-3 looks good.

image


https://github.com/sot/aca_stats/blob/master/starcheck_acq_probs_2015.ipynb

taldcroft: > I haven't been able to find any catalogs that exceed it (as in A products) to have a good before/after test.

This is where we need to get outside of pure regression testing with available products.  You can modify a backstop file to effectively remove stars.  Because of the connectivity with other products (guide summary) you might get other errors popping but, but you could at least do one functional test (even if it doesn't get formally incorporated into testing).  I'm hoping to transition to starhopper within a year.

jeanconn:
Todo:

- [ ] Add functional test of CUM_PROB_LIMIT

taldcroft:
As mentioned, I'm OK with any sort of on-the-side test with documentation in the PR that says it passed.

src/lib/Ska/Starcheck/FigureOfMerit.pm
taldcroft: Can we just turn that line in the existing summary to red?  It's already likely that anything which fails this test will be red or at least obvious at the summary level, so I'm not sure we need a new report element here.

jeanconn: I don't have per-obsid full-line colorization in the summary.  We just have
the count of red and green warnings.

This is also the only way changes to this pop up in the regression test
diffs.

jeanconn: And I should have said "add a warning to the list of per-obsid warnings" instead of "add a warning to the block" in the original line note for this.

jeanconn:
For on-the-side testing, I can comment that if I set the magnitudes to 12 for the first 5 stars in the first obsid in the AUG0104 test products, I see the warning new warning when we exceed the probability limit and the text turns red.  

bad_probs

The magnitudes here were set using the guide summary and not backstop, so the MAXMAGs weren't adjusted accordingly, and the newly dim stars now show warnings about spoilers etc.

taldcroft:
Looks good!

jeanconn:
There are no open actions remaining on this PR, so I think this is good to go.

PR #119, opened 2015-07-18
Use Python for acq probs and replace MINMAG with P_SUCC in report [Merge before #117]

taldcroft:
This is based on the star-probs branch.  Sample starcheck report output is:
```
---------------------------------------------------------------------------------------------
 IDX SLOT        ID  TYPE   SZ  P_SUCC    MAG   MAXMAG   YANG   ZANG DIM RES HALFW PASS NOTES
---------------------------------------------------------------------------------------------
[ 1]  0           1   FID  8x8   1.000   7.000   8.000    940  -1722   1   1   25          
[ 2]  1           4   FID  8x8   1.000   7.000   8.000   2157    181   1   1   25          
[ 3]  2           5   FID  8x8   1.000   7.000   8.000  -1808    175   1   1   25          
[ 4]  3  1032856344   BOT  6x6   0.946   9.342  10.844  -1149   -864  20   1  120          
[ 5]  4  1032858888   BOT  6x6   0.902   9.463  10.969  -1685   2286  16   1  100  a2g2    
[ 6]  5  1032860872   BOT  6x6   0.972   9.221  10.719  -1669   -312  20   1  120    a2    
[ 7]  6  1032862008   BOT  6x6   0.984   9.119  10.625  -1005   -308  20   1  120          
[ 8]  7  1032867928   BOT  6x6   0.985   8.353   9.859  -1983  -1396  20   1  120          
[ 9]  0  1032850904   ACQ  6x6   0.604   9.836  11.344  -1766  -2105  10   1   70    a2    
[10]  1  1032873416   ACQ  6x6   0.887   9.493  11.000   -640   1667  20   1  120    a2    
[11]  2  1032868064   ACQ  6x6   0.823   9.600  11.109   2045   1823  20   1  120    a2    

Probability of acquiring 2,3, and 4 or fewer stars (10^x):	-5.344	-3.767	-2.470	
Acquisition Stars Expected  : 7.10
Predicted Max CCD temperature: -15.6 C 	 N100 Warm Pix Frac 0.051 
```


jeanconn:
This will probably need corresponding updates to StarcheckParser.pm and mica.starcheck.parser.

src/starcheck.pl
```diff
@@ -51,13 +51,16 @@
 import perl
 # this provides an interface back to the Perl namespace
 
+try:
```
jeanconn: I *thought* I'd stuck the try block in ccd_temp_wrapper because the perl warning array wasn't defined yet if this Inline piece was evaluated first... or something like that.

src/lib/Ska/Starcheck/FigureOfMerit.pm
```diff
@@ -28,42 +24,63 @@ our @EXPORT = qw();
...
+use Inline Python => q{
+import sys
+import perl
+# this provides an interface back to the Perl namespace
+
+try:
+    from starcheck.star_probs import acq_success_prob, prob_n_acq
+except ImportError as err:
+    # write errors to starcheck's global warnings and STDERR
+    perl.warning("Error with Inline::Python imports \n".format(err))
```
jeanconn: warning() is that function in starcheck.pl in the original use of this code, and warning() isn't defined in FigureOfMerit.pm for this to do anything.  Do we want to try to catch import errors in general at this point or should starcheck just exit if we have trouble loading Python components?

src/lib/Ska/Starcheck/FigureOfMerit.pm
```diff
@@ -28,42 +24,66 @@ our @EXPORT = qw();
 our @EXPORT_OK = qw( make_figure_of_merit );
 our %EXPORT_TAGS = ( all => \@EXPORT_OK );
 
+use Inline Python => q{
+import sys
+import perl
+# this provides an interface back to the Perl namespace
+
+try:
+    from starcheck.star_probs import acq_success_prob, prob_n_acq, mag_for_p_acq
```
jeanconn: Is mag_for_p_acq going into chandra_aca with the rest or staying in starcheck?

src/lib/Ska/Starcheck/FigureOfMerit.pm
```diff
@@ -28,42 +24,66 @@ our @EXPORT = qw();
...
 
 sub make_figure_of_merit{
     my $c;
     my $self = shift;
     return unless ($c = $self->find_command("MP_STARCAT"));
+
     my @probs;
     my %slot_probs;
+
+    my $t_ccd = 19.0;  # ??
```
jeanconn: Not sure about the ?? (work in progress) but I think you want $self->{ccd_temp} at this point.  I can also rename all of the ccd_temp pieces to t_ccd now that I know that's your preferred labeling.  Do you want a default value if, for any reason, the temperature could not be determined from xija?  If so, should that be set here, or set in set_ccd_temp() in Ska::Starcheck::Obsid?  If a default is desired, I think that -14 might be reasonable.

src/lib/Ska/Starcheck/FigureOfMerit.pm
```diff
@@ -28,42 +24,66 @@ our @EXPORT = qw();
...
+    # Calculate the probability of acquiring n stars
+    my ($n_acq_probs, $n_or_fewer_probs) = _prob_n_acq(\@probs);
+
+    $self->{figure_of_merit} = {expected => substr(sum(@probs), 0, 4),
+                                cum_prob => [map { log($_) / log(10.0) } @{$n_or_fewer_probs}],
+                                cum_prob_bad => ($n_or_fewer_probs->[2] > $CUM_PROB_LIMIT)
+                                };
 
+    # Dynamic mag limits based on 75% and 50% chance of successful star acq
+    $self->{mag_faint_yellow} = mag_for_p_acq(0.75, $date, $t_ccd);
```
jeanconn: I think these need to be in their own routine.  check_star_catalog() needs the limits to do its mag limit checks.  make_figure_of_merit() uses the output of the spoiler checks from check_star_catalog to determine the probabilities, so in the current order, make_figure_of_merit comes after check_star_catalog.  Thus the easiest thing to do would probably be a separate routine to get the mag limits.

src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -921,6 +920,9 @@ sub check_star_catalog {
     my $min_fid      = 3;
     ########################################################################
 
+    $self->{mag_faint_red} = 15.0;
+    $self->{mag_faint_yellow} = 15.0;
```
jeanconn: ?

src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1124,10 +1126,10 @@ sub check_star_catalog {
 	# Faint and bright limits ~ACA-009 ACA-010
 	if ($type ne 'MON' and $mag ne '---') {
 
-	    if ($mag < $mag_bright or $mag > $mag_faint_red) {
+	    if ($mag < $mag_bright or $mag > $self->{mag_faint_red}) {
```
jeanconn: Did we decide in the working group to write out the mag limit that was used in the warning?  Or to put it in the block that includes the probabilities.

src/lib/Ska/Starcheck/FigureOfMerit.pm
taldcroft: Yes, https://github.com/sot/chandra_aca/pull/5

src/lib/Ska/Starcheck/FigureOfMerit.pm
taldcroft: We should consider the temperature predictions to be a requirement of running ACA review.

src/lib/Ska/Starcheck/FigureOfMerit.pm
taldcroft: OK sounds reasonable.

taldcroft:
The whole `star_probs.py` module should be deleted.

src/lib/Ska/Starcheck/Obsid.pm
taldcroft: I think that we decided to put in the probabilities block (or somewhere persistent anyway).  That way you always know the values even if there are no warnings.

src/lib/Ska/Starcheck/Obsid.pm
taldcroft: Exactly.  That should just be deleted, but I'm not sure where that was going.

src/lib/Ska/Starcheck/FigureOfMerit.pm
jeanconn: Even when considered a requirement of review, I don't want the software to just die if run on re-open products instead of interrupt products (as in on the AUG1407A products in the regression test suite).

jeanconn: But we probably also don't need to call figure of merit on the obsids that are incomplete or a mess.

src/lib/Ska/Starcheck/Obsid.pm
jeanconn: I had some text in
https://github.com/sot/starcheck/commit/e463a46992a58e681175198a448b3e7e2a0c554c
that should work.

src/lib/Ska/Starcheck/FigureOfMerit.pm
taldcroft: I think these limits should be capped at 10.6 (red) and 10.3 (yellow).  I noticed some large values (> 11 mag) in the regression diffs for old catalogs.

jeanconn: That makes sense.  I was originally thinking about only applying the dynamic limits after a time T to avoid the non-informative regress diffs but the functional change right here would be as effective and less code messy.

PR #123, opened 2015-11-12
Check date of characteristics file

taldcroft:
Check that the characteristics file date is not more than 30 days before the first load command.

### DIFF
--- a/src/starcheck.pl
+++ b/src/starcheck.pl
@@ -51,7 +51,7 @@ import sys
 import perl
 # this provides an interface back to the Perl namespace
 
-from starcheck.pcad_att_check import make_pcad_attitude_check_report
+from starcheck.pcad_att_check import make_pcad_attitude_check_report, check_characteristics_date
 
 def ccd_temp_wrapper(kwargs):
     try:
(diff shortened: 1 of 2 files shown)


src/starcheck.pl
```diff
@@ -678,10 +678,16 @@ sub json_obsids{
...
         if ($att_ok){
-            $out .= "<A HREF=\"${att_report}\">[OK] Coordinates as expected.</A>\n\n";
+            $out .= "<A HREF=\"${att_report}\">[OK] Coordinates as expected.</A>\n";
         }
         else{
-            $out .= "<A HREF=\"${att_report}\">[NOT OK] Coordinate mismatch or error.</A>\n\n";
+            $out .= "<A HREF=\"${att_report}\">[NOT OK] Coordinate mismatch or error.</A>\n";
+        }
+        if (check_characteristics_date($char_file, $date[0])){
+            $out .= "<A HREF=\"${att_report}\">[OK] Characteristics file newer than 30 days</A>\n\n";
```
jeanconn: For extra credit, should this be a relative link to the characteristics file used (instead of a link to the list of obsid checks)?

jeanconn: Nah, that wouldn't work in all cases where starcheck is called from outside of the products directory.  So we'd need to copy the characteristics into $STARCHECK and I suppose that would increase the risk that they'd get copied somewhere they shouldn't.

taldcroft: Good catch, but technically the characteristics file is not available as a relative link since it would need to be placed in the starcheck output directory.

PR #124, opened 2015-11-13
Hopper continuity

jeanconn:
Parse SIM history for hopper initial state continuity.

### DIFF
--- a/starcheck/pcad_att_check.py
+++ b/starcheck/pcad_att_check.py
@@ -1,6 +1,6 @@
 import re
 import hopper
-from parse_cm import read_maneuver_summary
+from parse_cm import read_backstop, read_maneuver_summary
 from Chandra.Time import DateTime
 
 
@@ -35,8 +35,18 @@ def test_check_characteristics_date():
                                     '2015Nov02 at 00:00:00.000')
     assert ok is False
 
+def recent_history(time, file):
+    for state_line in reversed(open(file).readlines()):
+        state_match = re.match('^(\d+\.\d+)\s+\|\s+(\S+)\s*$',
+                                state_line)
+        if state_match:
+            if (DateTime(state_match.group(1), format='greta').secs
+                < time):
+                return state_match.group(1), int(state_match.group(2))
+
 
 def make_pcad_attitude_check_report(backstop_file, or_list_file=None, mm_file=None,
+                                    simtrans_file=None, simfocus_file=None,
                                     ofls_characteristics_file=None, out=None,
                                     ):
     """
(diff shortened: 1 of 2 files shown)


jeanconn:
Not sure why 4e1af82 and bf0ac61 are showing up in the diffs at this point.

jeanconn:
This could use more checking, but seems to fix the NOV1615 problem where 17677 is checked against HRC-I SI-ALIGN instead of ACIS-I.  Also, not sure if it might have made sense to do the hopper check with the OR detector instead of the hopper determined detector.

taldcroft:
I don't understand the extra commits here either. I just changed the commit message and force pushed, which fixed things up.

starcheck/pcad_att_check.py
```diff
@@ -35,8 +35,18 @@ def test_check_characteristics_date():
...
+        state_match = re.match('^(\d+\.\d+)\s+\|\s+(\S+)\s*$',
+                                state_line)
+        if state_match:
+            if (DateTime(state_match.group(1), format='greta').secs
+                < time):
+                return state_match.group(1), int(state_match.group(2))
+
 
 def make_pcad_attitude_check_report(backstop_file, or_list_file=None, mm_file=None,
+                                    simtrans_file=None, simfocus_file=None,
```
taldcroft: If this file list expands any more we should call it `files` and make it a dict like `{'backstop': .., 'or_list': .., ..}`.  OK for now.

starcheck/pcad_att_check.py
```diff
@@ -35,8 +35,18 @@ def test_check_characteristics_date():
                                     '2015Nov02 at 00:00:00.000')
     assert ok is False
 
+def recent_history(time, file):
+    for state_line in reversed(open(file).readlines()):
```
taldcroft: In this small routine `state_` everywhere is just adding noise because it isn't helping to understand what is in the line (yes, a "state", but what is that?).  It makes the code harder to read because the variable names are more difficult to visually distinguish (`state_match` vs. `state_line`).  

Let's shoot for a consistent pattern of using `line` for a line in a file or list of lines, and `match` for a regex match, `ok` for a one- or two-off boolean check, and putting a comment line describing things.  (And of course don't use this pattern when it really doesn't make sense, perhaps when the context is more than a screenfull of code lines).

starcheck/pcad_att_check.py
```diff
@@ -45,18 +55,20 @@ def make_pcad_attitude_check_report(backstop_file, or_list_file=None, mm_file=No
...
     q = [mm[0][key] for key in ['q1_0', 'q2_0', 'q3_0', 'q4_0']]
+    bs = read_backstop(backstop_file)
+    simfa_time, simfa = recent_history(DateTime(bs[0]['date']).secs,
+                                       simfocus_file)
+    simpos_time, simpos = recent_history(DateTime(bs[0]['date']).secs,
+                                       simtrans_file)
     initial_state = {'q_att': q,
-                     'simpos': 75624,
-                     'simfa_pos': -468}
-
```
taldcroft: Put back this whitespace.

starcheck/pcad_att_check.py
```diff
@@ -35,8 +35,18 @@ def test_check_characteristics_date():
                                     '2015Nov02 at 00:00:00.000')
     assert ok is False
 
+def recent_history(time, file):
+    for state_line in reversed(open(file).readlines()):
+        state_match = re.match('^(\d+\.\d+)\s+\|\s+(\S+)\s*$',
+                                state_line)
+        if state_match:
+            if (DateTime(state_match.group(1), format='greta').secs
```
taldcroft: Split out the fields here:
```
greta_time, value = match.groups()
if DateTime(greta_time, format='greta').secs < time:
    return greta_time, int(value)
```

starcheck/pcad_att_check.py
taldcroft: Good idea with the reverse indexing of the file lines.

taldcroft:
Apart from style comments this looks good!

taldcroft:
:+1:

starcheck/pcad_att_check.py
jeanconn: I don't see it expanding more for this check, but I figured in general you'd soon move to just fetching files for hopper using a template.

PR #126, opened 2015-11-23
Continued dynamic limit work [Merge before #119]

jeanconn:
Integrate dynamic limits


src/starcheck.pl
```diff
@@ -578,6 +566,7 @@ sub json_obsids{
 
     $obs{$obsid}->check_monitor_commanding(\@bs, $or{$obsid});
     $obs{$obsid}->check_flick_pix_mon();
+    $obs{$obsid}->set_dynamic_mag_limits();
```
jeanconn: Do we want a start date for the use of the dynamic mag limits in the code?  There could also be an option to turn off the dynamic limits... Mostly I'm wondering about the regression test set.

taldcroft: I think that's a slippery slope that could apply to every change that affects the output format.  Does the code actually run OK on very old loads?  I think the previous PCAD alignment check update was a little different than this because of the underlying process and interface change that drove the new check.

taldcroft:
I haven't traced this through in detail, but at first glance it looks like `mag_faint_red` (and yellow) won't ever get set if there is no `self->{t_ccd}`.  Is that right?  Will that be a problem?  Is the case where the CCD temp is not available actually tested?

One idea could be to set the CCD temp to the MP guideline value if not available.  (And again, with the current code is that actually possible?)  This basically gives the worst case number but let's all the computations go through.

src/starcheck.pl
jeanconn: Running it on very old loads is/should be fine.  I was doing that anyway to run the code through its paces as that is the primary goal of the regression tests.  I just wasn't sure if we wanted to see all of the old values change in the regression test output.  I'll continue with that as that plan.

jeanconn:
I believe the case where the CCD temp is not available is presently tested via the re-open products in the regression tests, so that should help.  Regarding default values, I've got the planning limit in the starcheck characteristics as "ccd_temp_red_limit: -14".  The Perl code has access to that value so we should be able to set that as the default if there's a problem.

jeanconn:
For the case where we have no CCD prediction, do we want to use the planning limit value for the mag limit determination *and* the figure of merit probabilities?  Or just the mag limits (without calculating or printing the figure of merit values)?

taldcroft:
Which choice is both reasonable and simplest?

jeanconn:
It is probably cleanest to just do the acquisition probability predictions at the planning limit temperature.  There'll be one red warning that there was no t_ccd prediction for the obsid.

taldcroft:
Bingo.

jeanconn:
Now updated with default ccd temperature, some new explicit casts to floats to keep the Python happy when the Perl does moderately weird things, and some new logic in the output printing.

See #127 for the latest diffs.

jeanconn:
Maximums set in 6b17bca and diffs redone.

PR #127, opened 2015-12-03
Update regress output [output for #126]

jeanconn:
Update regress output for dynamic mag limits, modern acq probabilities
in figure of merit output, and new P_SUCC column in catalog data.


jeanconn:
Diffs in this PR are not needed.

PR #98, opened 2015-12-03
Change not enough bright stars in perigee to info statement for special case ERs

jeanconn:
Special case ERs should not get the "too few bright stars in perigee" warning.
Change to info statement.


taldcroft:
:+1:

PR #95, opened 2015-12-04
Flag non-standard dither and apply monitor window checks as needed

jeanconn:
Obs 17383 in FEB2315A has a non-standard dither with a 64 arcsec amplitude in Z.

From Eric:

In order to have these dither parameters, dither commanding must be similar to that for observations with monitor windows.
 
Dither (using standard parameters) must be disabled 1 min prior to the end of the maneuver to the observation that will use the large dither parameters.  5 min after the end of the maneuver (when all stars should have been acquired and are being tracked), dither (using the large parameters) can be re-enabled.  Finally, 5 min prior to the end of the observation, dither must be commanded back to standard parameters.

### DIFF
--- a/src/lib/Ska/Starcheck/Obsid.pm
+++ b/src/lib/Ska/Starcheck/Obsid.pm
@@ -610,7 +610,6 @@ sub check_dither {
     my $self = shift;
 
     my $dthr = shift;		  # Ref to array of hashes containing dither states
-    my $backstop = shift;
     my %dthr_cmd = (ON => 'ENAB',   # Translation from OR terminology to dither state term.
 		    OFF => 'DISA');
 
--- a/src/lib/Ska/Starcheck/Obsid.pm
+++ b/src/lib/Ska/Starcheck/Obsid.pm
@@ -624,6 +624,10 @@ sub check_dither {
     if ( $self->{obsid} =~ /^\d*$/){
 	return if ($self->{obsid} > 50000); # For eng obs, don't have OR to specify dither
     }
+    # If there's no starcat on purpose, return
+    if (defined $self->{ok_no_starcat} and $self->{ok_no_starcat}){
+        return;
+    }
     unless ($manvr = find_command($self, "MP_TARGQUAT", -1)
             and defined $manvr->{tstart}) {
 	push @{$self->{warn}}, "$alarm Dither status not checked\n";
--- a/src/lib/Ska/Starcheck/Obsid.pm
+++ b/src/lib/Ska/Starcheck/Obsid.pm
@@ -760,10 +760,11 @@ sub large_dither_checks {
     my $all_dither = shift;
     my $time_tol = 10;         # Commands must be within $time_tol of expectation
 
-    # Correct times in the maneuver summary by 10 seconds
-    my $manvr_sum_offset = 10;
-    my $obs_tstart = $self->{obs_tstart} - $manvr_sum_offset;
-    my $obs_tstop = $self->{obs_tstop} - $manvr_sum_offset;
+    # The obs tstart and tstop times are derived from the OFLS maneuver summary file,
+    # which for implementation reasons gives the maneuver times offset by 10 seconds
+    # from the actual command timing. Fix this here.
+    my $obs_tstart = $self->{obs_tstart} - 10;
+    my $obs_tstop = $self->{obs_tstop} - 10;
 
     # Now check in backstop commands for :
     #  Dither is disabled (AODSDITH) 1 min prior to the end of the maneuver (EOM)
--- a/src/lib/Ska/Starcheck/Obsid.pm
+++ b/src/lib/Ska/Starcheck/Obsid.pm
@@ -683,18 +683,20 @@ sub check_dither {
         }
     }
 
-
+    # Loop again to check for dither changes during the observation
+    # ACA-003
     if (not defined $obs_tstop ){
         push @{$self->{warn}},
             "$alarm Unable to determine obs tstop; could not check for dither changes during obs\n";
     }
     else{
-        # Loop again to check for dither changes during the observation
-        # ACA-003
-        foreach my $dither (@{$dthr}) {
+        foreach my $dither (reverse @{$dthr}) {
             if ($dither->{time} > ($obs_tstart + $obs_beg_pad)
                     && $dither->{time} <= $obs_tstop - $obs_end_pad) {
                 push @{$self->{warn}}, "$alarm Dither commanding at $dither->{time}.  During observation.\n";
+            }
+            if ($dither->{time} < $obs_tstart){
+                last;
             }
         }
     }
--- a/src/lib/Ska/Starcheck/Obsid.pm
+++ b/src/lib/Ska/Starcheck/Obsid.pm
@@ -614,10 +614,6 @@ sub check_dither {
     my %dthr_cmd = (ON => 'ENAB',   # Translation from OR terminology to dither state term.
 		    OFF => 'DISA');
 
-    my %standard_y_dither = (20 => 1087.0,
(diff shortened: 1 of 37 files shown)


jeanconn:
Code attached.

taldcroft:
I guess we rely on the new regression weeks for functional validation, though the coverage is not 100% (e.g. I don't think there is an example of the dither period being unexpected).

Hopper is coming along nicely and will support actual functional tests by allowing small backstop snippets to be run, including straightforward injection of mistakes like a wrong dither period.  In the very near future we should be thinking hopper for all new starcheck development.

jeanconn:
We still need to get a dither history file that has amplitude and period.

taldcroft:
We could always start down the path of relying on command states where History data is not available.  We already do this for thermal checking, though that is a bit squishier.

taldcroft:
(That of course also requires putting this into command states, but it will be a snap with dynamic cmd states).

jeanconn:
Certainly doable from cmd_states but this seems like a trivial initial
condition that should just be in a new dither history file.  SAUSAGE needs
it too, and I think we started to ask for it 4 years ago.​

taldcroft:
FYI here is the implementation and testing for dither state handling in hopper (with a little bit of extra cruft thrown in):
https://github.com/sot/hopper/commit/c2a1a65781491efdb5d19731d8d4601badf4e5dc

Notice how the testing can be really specific and localized.

jeanconn:
Is you implementation idea that we'll continue to develop hopper tests and use them in Perl starcheck, or develop on the side until you do a full switch to hopper?

taldcroft:
BTW I improved and extended the implementation in hopper: https://github.com/sot/hopper/commit/afa397a4b22c475817ecb1c94084e58c7ca83f7d.

I would say that going forward we should be thinking of Python and hopper for any new starcheck tests, where this makes sense.  Basically we don't want to duplicate code.  Currently the API in hopper is stabilizing but there are likely to be tweaks as things go forward. In fact actually using hopper for starcheck tests will be a good way to identify weaknesses or issues with the API.

In this case this PR with perl code should be used (after review is finished).

jeanconn:
Higher up in the conversation I had that question about AOACRSET.  Eric signed off on products without that command, so a check for it is not included in the extra "large dither" checks in this PR.

src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -682,6 +713,63 @@ sub check_dither {
...
+    my $t_manv = $manv->{tstop};
+    my %dt = (AODSDITH => -60, AOENDITH => 300);
+    my %cnt = map { $_ => 0 } keys %dt;
+    my $all_dither_count = 0;
+    foreach my $bs (grep { $_->{cmd} eq 'COMMAND_SW' } @{$backstop}) {
+       my %param = Ska::Parse_CM_File::parse_params($bs->{params});
+       next unless $bs->{time} >= ($t_manv - ($dt{AODSDITH} + $time_tol + 1));
+       next unless ($param{TLMSID} =~ /^AO/);
+       # Search for all relevant dither commands
+       foreach my $cmd ('AODSDITH', 'AOENDITH', 'AODITPAR') {
```
jeanconn: There is no check here for a return to standard parameters, so I think AODITPAR is a bit of a decoy.

src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -619,9 +619,16 @@ sub check_dither {
     my $self = shift;
 
     my $dthr = shift;		  # Ref to array of hashes containing dither states
+    my $backstop = shift;
     my %dthr_cmd = (ON => 'ENAB',   # Translation from OR terminology to dither state term.
 		    OFF => 'DISA');
 
+    my %standard_y_dither = (20 => 1087.0,
```
taldcroft: Need to include the other dither cases as well, probably encoded in a table like the one Erik provided.  That's my plan for hopper anyway.
```
Dither Case        Y Amplitude  Z Amplitude  Y Period  Z Period
-----------------  -----------  -----------  --------  --------
Standard ACIS           8.0          8.0      1000.0     707.1
Standard HRC           20.0         20.0      1087.0     768.6
Crab Observations       1.0          1.0      1000.0     707.1
ACIS Subarray           8.0          4.0      1000.0     707.4
ACIS Contaminant*       8.0         64.0      1000.0    2647.0
ACA DCC DFC             0.0          0.0      1000.0     707.1
```

jeanconn: I was under the impression/of the opinion that flagging all but the first two as  "non-standard" would be fine for this PR and that the rest would belong more in #96

taldcroft: I see your point.  Sounds like a good question for SSAWG.

jeanconn: Sounded like no real objections from SSAWG on this approach.

src/lib/Ska/Starcheck/Obsid.pm
jeanconn: Add a check for a return to standard parameters.

taldcroft: As mentioned today, from a system-level perspective the ACA checker should be responsible for making sure the dither parameters match the OR list, while the PCAD checker ensures that commanding is correct and safe for the commanded dither parameters.  Which is a way of saying you don't *have* to check for the return to standard parameters (and can spend effort elsewhere).

jeanconn: But, if we have already checked 2/3 of the things that we don't need to check, I'd be reassured if we just checked this too.

taldcroft: The "5 minutes before next maneuver" part of the requirement makes me think this won't be so easy to implement and will have corner case code.  If you are sure you can complete and fully test this new code in ~15 minutes then go for it, but I don't think there is any value added.

taldcroft: As for "reassured", I have 100% confidence that Larry will do the backstop check correctly.

jeanconn: My notes had 5 minutes before the end of the observation, which I was planning to just do via the or_er_tstop that should already be hanging around in the data structure for each obsid.

jeanconn:
I'm also of course fine with chucking the disable/enable checks; I was just still working from your comment that  "I think we do always want these checks for the existing non-standard dither configurations,".

src/lib/Ska/Starcheck/Obsid.pm
taldcroft: Eric's notes say "- Return to standard ACIS parameters 5 min before start of next maneuver ...".  Not sure when those might be different.

jeanconn: Gotcha.  I think we have two sets of Eric notes, because you opened this issue with the "Finally, 5 min prior to the end of the observation, dither must be commanded back to standard parameters."  But I think those are basically only different in the last observation in products.  For the last observation, I just use the "processing summary" stop time for the other checks (dither, momentum dumps).

jeanconn:
Code attached to make the large dither check at end of observation more clear.  This based on the comments that ended up in #131.  Does this help @taldcroft ?

jeanconn:
Now that you've given a :+1: , may I make one more change?  I had added a products that fail the large dither checks, but it would be good to add the B products too.

taldcroft:
ok.

jeanconn:
A few more updates responding to diffs in regression outputs.

taldcroft:
:+1: on new commits.

PR #131, opened 2015-12-14
Check catalogs using only dither from backstop or history

jeanconn:
Check catalogs using only dither from backstop or history.  

This still compares backstop dither state to OR state and compares OR vs backstop amplitudes if backstop commanding includes amplitude.  However, instead of using the OR amplitude by default in check_catalog(), the default value of 20 arcsecs will be used unless an AODITPAR has occurred to set dither parameters explicitly in backstop.

### DIFF
--- a/src/lib/Ska/Starcheck/Obsid.pm
+++ b/src/lib/Ska/Starcheck/Obsid.pm
@@ -21,6 +21,7 @@ package Ska::Starcheck::Obsid;
 use strict;
 use warnings;
 
+use List::Util qw( max );
 use Quat;
 use Ska::ACACoordConvert;
 use File::Basename;
@@ -1077,8 +1078,7 @@ sub check_star_catalog {
 
     my $dither;			# Global dither for observation
     if (defined $self->{cmd_dither_y_amp} and defined $self->{cmd_dither_z_amp}) {
-	$dither = ($self->{cmd_dither_y_amy} > $self->{cmd_dither_z_amp} ?
-		   $self->{cmd_dither_y_amp} : $self->{cmd_dither_z_amp});
+	$dither = max($self->{cmd_dither_y_amy}, $self->{cmd_dither_z_amp});
     } else {
 	$dither = 20.0;
     }
--- a/src/lib/Ska/Starcheck/Obsid.pm
+++ b/src/lib/Ska/Starcheck/Obsid.pm
@@ -1080,7 +1080,6 @@ sub check_star_catalog {
 	$dither = ($self->{cmd_dither_y_amy} > $self->{cmd_dither_z_amp} ?
 		   $self->{cmd_dither_y_amp} : $self->{cmd_dither_z_amp});
     } else {
-        $self->{used_default_dither} = 1;
 	$dither = 20.0;
     }
 
--- a/src/lib/Ska/Starcheck/Obsid.pm
+++ b/src/lib/Ska/Starcheck/Obsid.pm
@@ -686,8 +686,6 @@ sub check_dither {
     if (($bs_val eq 'ENAB') and (defined $dither->{ampl_y} and defined $dither->{ampl_p})){
         $self->{cmd_dither_y_amp} = $dither->{ampl_y};
         $self->{cmd_dither_z_amp} = $dither->{ampl_p};
-        $self->{cmd_dither_y_period} = $dither->{period_y};
-        $self->{cmd_dither_z_period} = $dither->{period_z};
         if (not standard_dither($dither)){
             push @{$self->{yellow_warn}}, "$alarm Non-standard dither\n";
             if ($dither->{ampl_y} > $large_dith_thresh or $dither->{ampl_p} > $large_dith_thresh){
--- a/src/lib/Ska/Starcheck/Obsid.pm
+++ b/src/lib/Ska/Starcheck/Obsid.pm
@@ -681,10 +681,12 @@ sub check_dither {
             push @{$self->{warn}}, $warn;
         }
     }
-
-
     # Check for standard and large dither based solely on backstop/history values
     if (($bs_val eq 'ENAB') and (defined $dither->{ampl_y} and defined $dither->{ampl_p})){
+        $self->{cmd_dither_y_amp} = $dither->{ampl_y};
+        $self->{cmd_dither_z_amp} = $dither->{ampl_p};
+        $self->{cmd_dither_y_period} = $dither->{period_y};
+        $self->{cmd_dither_z_period} = $dither->{period_z};
         if (not standard_dither($dither)){
             push @{$self->{yellow_warn}}, "$alarm Non-standard dither\n";
             if ($dither->{ampl_y} > $large_dith_thresh or $dither->{ampl_p} > $large_dith_thresh){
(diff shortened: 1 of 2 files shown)


jeanconn:
This depends on PR #95

src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -646,41 +648,163 @@ sub check_dither {
...
+    if ((abs($self->{obs_tstart} - $obs_start_dither->{time} - 60) > $time_tol)
+            or ($obs_start_dither->{state} ne 'DISA')){
+        push @{$self->{warn}},
+            sprintf("$alarm Dither should be disabled 1 min before obs start for Large Dither\n");
+    }
+
+    # Are parameters set to normal values 5 minutes before the end of the observation?
+    my $obs_stop_dither;
+    foreach my $dither (reverse @{$all_dither}) {
+	if ($self->{obs_tstop} - 300 >= $dither->{time}) {
```
taldcroft: This appears to be checking for any dither commanding that happens earlier than 5 mins before `obs_tstop`, not for commanding that occurs within some narrow window.  Maybe in conjunction with the logic of other tests that is sufficient, but it makes the tests more difficult to verify analytically.

jeanconn: The combination of reverse and last over the dither states finds the first state before the 5 minute mark and quits.  However, if, like the dither disable test, the real intent was to confirm that dither was set to nominal parameters and not changed through end of observation, I could just check the dither state at the end of the observation and confirm that it started 5 minutes before the end of the observation.  I was trying to take a bigger picture view here and test the thing that the the dither commanding was supposed to effect, and leave the PCAD check to search for specific commands at specific times.

taldcroft: You at least need some timing margin here.  I think that if the dither command occurs at 299.999 sec before `obs_tstop` then the test will fail.

jeanconn: I would think that should actually be a fail (in that greater than would be OK but less than 5 minutes would be bad.  But I'll change this to just a a tolerance around 300.  Granted, the tolerance really just seems to be capturing that the maneuver summary is off by 10 seconds.  Have I mentioned I won't miss this (when we're working more in starhopper)?

jeanconn:
Also, this PR was really just the last commit for the "use backstop dither"  

I see you've got comments on the large dither checks that are on the other PR.  The "use backstop dither" thing was chained to the end, but this time I just made it a full PR against master and figured I'd rebase once the large dither check PR was good to go.  Now just confused.

taldcroft:
I think I would have just called #95 "Enhance dither checking" and put this commit into it.   Basically all that code seems to be intertwined with this so just put in the same PR (e.g. next time).  For now this looks good enough code wise.

What I can't obviously see are any tests that show it works and has reasonable coverage.  I expect there are regression diffs somewhere but I don't specifically know where to look.

jeanconn:
> "Basically all that code seems to be intertwined with this so just put in
the same PR (e.g. next time). "

Disagreeing just in the fact that this change to use the backstop code is a
very different conceptual change and related to a different issue.  I
didn't want to try to deal with the complex merge if I had put it in a
separate branch off master, but I *did* want the idea independently
evaluated for merit to see if we were still going to do it.  I was planning
on it as a Y/N after review of the handle_big_dither changes.

taldcroft:
I agree that on further consideration this is independent.  However, it looks like a pretty self-contained change that should easily merge.  If you cannot easily merge some new feature then that's a good sign that it is intertwined with another changeset.  

In general rebasing and merging is possible, so if you have a conceptually distinct change that does not *explicitly* depend on functionality from another PR, then just start clean from master.

taldcroft:
Hmm, needs rebase now.

src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -681,10 +682,10 @@ sub check_dither {
             push @{$self->{warn}}, $warn;
         }
     }
-
-
     # Check for standard and large dither based solely on backstop/history values
     if (($bs_val eq 'ENAB') and (defined $dither->{ampl_y} and defined $dither->{ampl_p})){
+        $self->{cmd_dither_y_amp} = $dither->{ampl_y};
+        $self->{cmd_dither_z_amp} = $dither->{ampl_p};
```
jeanconn: I think it is appropriate and makes sense to only save out these amplitudes if dither is enabled (which is what this is doing).

taldcroft:
Looks good.  There is of course the option of using the OR value if it hasn't been commanded in the loads yet, but in the end we really don't *know* that the dither will match the OR value so I think it's better to be conservative here.  Was that your thinking?

jeanconn:
That was the driver and the thinking behind the original issue, but I don't
know if we want to run with it or not.  If we want to just use the OR
value, we don't need to apply this PR.  If we want to use the backstop
value only if the amplitude is defined (if there has been dither commanding
in the loads), we could do that too.

I think that the more conservative choice here (20 arcsecs) is fine from a
star catalog checking perspective, and if we really need the 12 arcsecs
margin, we should push to get the continuity right.
​

taldcroft:
Agreed (with your final sentence).

src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1074,9 +1075,8 @@ sub check_star_catalog {
     my $min_z = $z_ang_max;
 
     my $dither;			# Global dither for observation
-    if (defined $self->{DITHER_Y_AMP} and defined $self->{DITHER_Z_AMP}) {
-	$dither = ($self->{DITHER_Y_AMP} > $self->{DITHER_Z_AMP} ?
-		   $self->{DITHER_Y_AMP} : $self->{DITHER_Z_AMP}) * 3600.0;
+    if (defined $self->{cmd_dither_y_amp} and defined $self->{cmd_dither_z_amp}) {
+	$dither = max($self->{cmd_dither_y_amy}, $self->{cmd_dither_z_amp});
```
jeanconn: In general, there is now the over simplification that the observation "commanded" dither in this case should be the dither state that is running 8 minutes after the last maneuver.  That is reasonable for the guide star checks, but for the large dither observations, should the acquisition checks be run against the acquisition dither?

taldcroft: Yes, that's correct.  The current result should be conservative, so I think that in most cases we can live with this.  If we get in trouble with spoilers then we'll see.  Do you have a simple plan in mind to implement the correct behavior?

jeanconn: No plan in mind.  If I were, say, developing a stateful checker, I'd stay aware of checking the two star catalogs with separate states and parameters.

taldcroft: Fortunately with hopper this kind of thing happens without even trying.  You always are checking against the state values that apply at that time.  :smile_cat:  But I'm not sure what you mean about *two* star catalogs.  Am I wrong in thinking that for acquisition there is only one acquisition (that gets done with "small" dither)?  The guide stars are acquired with small dither as well, and then dither gets changed.  So that would imply guide stars should be checked at some point (perhaps 8 minutes) into the observation when the "observation" dither is in place.

jeanconn: I'm not sure if there are two acquisitions in the MON case or large dither setup.  I'll look at that again.  In general, by two catalogs, I just meant independently checking the acquisition stars and the guide/fid sets as two different 8-object sets.

jeanconn: Two acquisitions in the MON case, but only one normal acquisition with disabled dither in the large dither setup.

jeanconn: Also, a bug there as I've got {cmd_dither_y_amy}  .  And it isn't like Perl would tell me...

jeanconn:
@taldcroft is this good to go from your perspective?  Diffs at #133

taldcroft:
:+1:

PR #136, opened 2016-01-15
Use setuptools not distutils

taldcroft:


jeanconn:
I note that starcheck has the astropy version magic.  This means if I try to install something with a dev version I get something like this:
```
The version specified (u'1.1dev-r590-3c6c136') is an invalid version 
```
And I end up installing starcheck with the sha as the version in `conda list`
```
dska-fido: conda list starcheck
# packages in environment at /data/fido/ska/arch/x86_64-linux_CentOS-5:
#
starcheck-1.1dev-r590     3c6c136                   <pip>
```
Is that what we want?  I'm not sure.

taldcroft:
What is the behavior with the version cast to `str`?  setuptools should really be able to handle unicode for `version`, but if not then we need to force the version magic to return a `str`.  I.e. fix the problem at the source.

jeanconn:
No change with explicit cast.  I think something like "'1.2dev-r590-3c6c136'" is already being handled as a string.

jeanconn:
(just as a unicode string if we don't use an explicit cast... right?)

taldcroft:
OK, if I understand the `str(version)` was not improving the situation, so the change introduced here should stay.  There is a separate issue of how conda is handling our odd-ish version tag.

jeanconn:
Yes.  I don't believe that the changes in this PR make it any worse, just was wondering if the point of the PR was to really change to setuptools or to fix the output of conda list.

taldcroft:
The PR was to change setuptools, but I couldn't resist cleaning up a useless cast to `str` (but I hadn't realized that the original is unicode).

jeanconn:
I think the original is not unicode for a release (no dev means no git string).

taldcroft:
Sorry, this was intended to fix the original problem with conda list, but not the one you just showed.  But I don't know what the original problem was (in detail), but I assume that *that* problem is fixed.

jeanconn:
Yes!  I can confirm that the change from distutils to setuptools does fix the issue where `conda list` was apparently reporting the first version with an egg directory or some such in site-packages instead of reporting the most recent installed version.

jeanconn:
:+1:

jeanconn:
PS I tested this in my dev ska by 

  * removing all starchecks from site-packages
  *  installing an old version
  * confirming that version matched conda list
  * installing an updated starcheck
  * confirming that conda list had the old version instead of the expected version
  * switching to this branch
  * installing again
  * confirming that version matched the expected version in conda list
  * incrementing the version and repeating the check.

taldcroft:
I'll let you merge assuming this is good for the upcoming release.

jeanconn:
Looks good for the release to me.

Issue #96, opened 2015-02-20
Add comments in summary about special observation features


jeanconn:
Per Eric's idea:

Specifically, it would be nice to see which observations have monitor windows ("MON") and which have momentum unloads ("NPM Unload" or "NMM Unload").
 
Also, when dark current measurements are present, the attitude labels could be included ("DFC_I", "DC_T0", "DFC_0", "DC_T1", ...).
 
These brief notes would show up after the "WARNINGS [nn]" note(s) on the line for each appropriate Obs ID.
 
Given the current schedule, we could also note "Large Dither" & "Crab Dither", since I think these are the only 2 non-standard parameter sets.

jeanconn:
Per @emartin496

jeanconn:
This is still hanging around on my list, but I note that because ORVDot has removed the DOT/OFLSids that look like DFC_I, DC_T0, etc, I need to update the dark cal checker to use the dark current replica commanding to identify these (#176).  So, we'd need to use the dark current checker to pass those identifications back to starcheck.pl to get that content in the summary.

PR #140, opened 2016-03-05
Handle warnings on magnitude a little differently from other warnings

jeanconn:
While magnitude warnings are likely to improve with a transition to probability models that are calibrated with the MS filter disabled, at this point we have lots of warnings related to Magnitude which pop up on the summary and are printed for each observation.  This PR removes single magnitude warnings from the summary counts but still includes them in the warning "penalties" on each star for the warning-free counts.  Additionally magnitude warnings are printed in a separate chunk for each observation.

At this point, we care more about a single red warning on the overall probability of the observation than we care about say 5 red magnitude warnings, so I think this PR improves signal to noise a bit.

### DIFF
--- a/src/lib/Ska/Starcheck/Obsid.pm
+++ b/src/lib/Ska/Starcheck/Obsid.pm
@@ -1958,21 +1958,42 @@ sub print_report {
     
     $o .= "\n" if (@{$self->{warn}} || @{$self->{yellow_warn}} || @{$self->{fyi}} );
 
+    my $mag_regex = qr/^$alarm \[\s?\d+\] Magnitude. \s+\d+\.\d+\n/;
+    my @mag_red = grep {/$mag_regex/} @{$self->{warn}};
+    my @non_mag_red = grep {!/$mag_regex/} @{$self->{warn}};
+    my @mag_yellow = grep {/$mag_regex/} @{$self->{yellow_warn}};
+    my @non_mag_yellow = grep {!/$mag_regex/} @{$self->{yellow_warn}};
 
-
-    if (@{$self->{warn}}) {
+    if (@mag_red){
 	$o .= "${red_font_start}";
-	foreach (@{$self->{warn}}) {
+	foreach (@mag_red) {
 	    $o .= $_;
 	}
-	$o .= "${font_stop}";
+        $o .= "${font_stop}";
     }
-    if (@{$self->{yellow_warn}}) {
+    if (@mag_yellow){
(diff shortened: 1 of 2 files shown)


jeanconn:
New:

feb2216b_thispr

Current:
feb2216b_release

jeanconn:
For each observation, the second commit in this PR prints the magnitude warnings in a separate chunk
mag_chunk

jeanconn:
@taldcroft So this was just a thought I had after the MAR0516 load review.  Do you agree that we're more interested in noticing one red warning on probability for the observation than 5 magnitude warnings for an observation, and if so, do you think removing them from the summary would be helpful for your starcheck review?  Or do you not look at the summary warnings and Eric and I should just figure out what works for us?

taldcroft:
@jeanconn - the concept looks good to me (haven't reviewed code), but you should definitely bring this to Eric's attention.

jeanconn:
Hi @emartin496 .  

What do you think of this idea for a change in starcheck's presentation?  It removes magnitude warnings on stars from the per-obsid warning counts in the summary section at the top.    I've got a couple of images on the github page for it that show how it would change the display.  

This isn't related to the more complicated and more useful changes you have requested for the summary ( #96 ), but filtering out magnitude warnings from the top stuff is a pretty simple code change and I think it would help me to see the more important warnings.

As I note in the top, the changes I suggest here still count magnitude warnings against stars for the "warning free count" in those columns.

jeanconn:
Comments from Eric:

This change is OK.
 
I look at the summary and check first each ObsID that has any warnings listed.  With the new scheme, some ObsIDs won't show up as having any warnings in the summary, but that's OK, since I'll just catch them on a later pass-through of all ObsIDs.
 
Separating the magnitude warnings into their own block makes it a little harder to see all warnings associated with a particular entry, but that change is OK, too.

jeanconn:
My reply and plan:

OK.  It sounds to me like this change wouldn't bother you much, but
also wouldn't help with your process much either.  I think I'll plan
to leave the change on the shelf, so to speak, until after we make a
full transition to both disabling the MS filter during tracking and
using the MS filter disabled starcheck acquisition model.  As I said
in the beginning of the github pull request, there should be fewer
magnitude warnings then, as the 50 and 75% mag warning thresholds
should become fainter.  If at that point I'm still a bit distracted by
magnitude warnings, I'll bring this code up again.

jeanconn:
Overtaken by events (#223 )

PR #142, opened 2016-04-10
Explicitly set MS filter enabled before and disabled after 2016:102

jeanconn:
Explicitly set MS filter enabled before and disabled after 2016:102

### DIFF
--- a/src/starcheck.pl
+++ b/src/starcheck.pl
@@ -331,16 +331,9 @@ Ska::Starcheck::Obsid::set_odb(%odb);
 
 Ska::Starcheck::Obsid::set_config($config_ref);
 
-my $MSF_ENABLED;
 # Set the multple star filter disabled in the model if after this date
-if ($bs[0]->{time} > date2time('2016:102:00:00:00.000')){
-    set_acq_model_ms_filter(0);
-    $MSF_ENABLED = 0;
-}
-else{
-    set_acq_model_ms_filter(1);
-    $MSF_ENABLED = 1;
-}
+my $MSF_ENABLED = $bs[0]->{time} > datetime('2016:102:00:00:00.000');
+set_acq_model_ms_filter($MSF_ENABLED);
 
 # Read Maneuver error file containing more accurate maneuver errors
 my @manerr;
@@ -432,7 +425,6 @@ my %guidesumm = Ska::Parse_CM_File::guide($guide_summ) if (defined $guide_summ);
 
 # After all commands have been added to each obsid, set some global
 # object parameters based on commands
-
 
 foreach my $obsid (@obsid_id) {
     $obs{$obsid}->set_obsid(\%guidesumm); # Commanded obsid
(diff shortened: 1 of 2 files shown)


src/starcheck.pl
```diff
@@ -331,8 +331,16 @@
 
 Ska::Starcheck::Obsid::set_config($config_ref);
 
-
-
+my $MSF_ENABLED;
```
taldcroft: DRY
```
my $MSF_ENABLED = ($bs[0]->{time} > date2time('2016:102:00:00:00.000'));
set_acq_model_ms_filter($MSF_ENABLED);
```

src/starcheck.pl
```diff
@@ -666,7 +675,14 @@ sub json_obsids{
 	$save_hash{run}{badpix} = $ACA_badpix_date;
     }
 
+    if ($MSF_ENABLED){
```
taldcroft: DRY
```
$out .= "Using acquisition model for multiple star filter " 
        . ($MSF_ENABLED ? "enabled\n" : "disabled\n");
```

src/starcheck.pl
taldcroft: I think you don't need the `date2time`:
```
$bs[0]->{date} > '2016:102:00:00:00.000'
```

jeanconn: It isn't strictly required, but if we use string comparison it needs to be

$bs[0]->{date} gt ''2016:102:00:00:00.000'

and I *think* there are locale specific sort orders when you get into
string comparisons, so the numerics might be safer.  Granted, we are
probably using that string comparison in other places in the code.

src/starcheck.pl
```diff
@@ -331,8 +331,9 @@
 
 Ska::Starcheck::Obsid::set_config($config_ref);
 
-
-
+# Set the multple star filter disabled in the model if after this date
+my $MSF_ENABLED = $bs[0]->{time} > datetime('2016:102:00:00:00.000');
+set_acq_model_ms_filter($MSF_ENABLED);
```
jeanconn: I don't have an easy way to set thi by keyword (pass ms_enabled=True/False) from Perl without using a wrapper, so is it acceptable to just set the value of ms enabled via arg directly like this?  Will that be an API problem later if you want to set other fit pars? 

Regarding the actual argument passed, there's also an Inline::Python object that can pass True/False, but that didn't seem to add any value over the 0/1 that goes through here.

src/starcheck.pl
```diff
@@ -331,8 +331,9 @@
 
 Ska::Starcheck::Obsid::set_config($config_ref);
 
-
-
+# Set the multple star filter disabled in the model if after this date
+my $MSF_ENABLED = $bs[0]->{time} > datetime('2016:102:00:00:00.000');
```
jeanconn: Acceptable date?  And OK to just leave in the code for this?

jeanconn:
This ended up merged with a bug.  Fixed in https://github.com/sot/starcheck/commit/cb82a81595b767e8f742e52b347cffcdc0c8e88e for reference.

PR #143, opened 2016-04-10
Explicitly set MS filter disabled

jeanconn:
Explicitly set MS filter disabled


jeanconn:
Closing in favor of #142

PR #144, opened 2016-04-10
Explicitly set MS filter from characteristics

jeanconn:
Explicitly set MS filter from characteristics.yaml file


jeanconn:
Closing in favor of #142

PR #148, opened 2016-05-23
Handle new DOT trace ids ending in MXXX instead of MTLB

jeanconn:
Handle new DOT trace ids ending in MXXX instead of MTLB


jeanconn:
It looks like github caught up and I could finally see the branch and make a PR against it.  This PR is informational-only at this point.

taldcroft:
You can guess the all-caps three letter comment I would make and save me from having to type it.  :smile:

jeanconn:
I had a silly idea that this reduced the diff, but agree now that makes no
sense.

taldcroft:
Apparently we agreed to this change a while back and so we will want to run with this change.

PR #150, opened 2016-05-26
Copy starcheck python files into working directory in regression test

jeanconn:
Copy starcheck python files into working directory in regression test

' ' is at the beginnign of sys.path, so copying over the top level python
files into '' before running each regression starcheck run gets around the
path issue that can arise when a starcheck egg dir in the environment is
before the PYTHONPATH used in the sandbox_starcheck testing.

### DIFF
--- a/src/run_regress
+++ b/src/run_regress
@@ -34,8 +34,10 @@ RunRegression()
     cd $test
 
 
-  # Run test version.  Use the 'starcheck' launcher to activate the dev environment
+  # Run test version.  Use the 'sandbox_starcheck' launcher to activate the dev environment
   #
+    echo "Copying python files for local use"
+    cp -Ruva ${home}/starcheck .
     echo "Running: ${home}/sandbox_starcheck -vehicle -agasc $agasc -fid_char $fid_char -dir $mphome/$load"
     echo "******************** (TEST VEHICLE) $load *******************" >> $vlog
     ${home}/sandbox_starcheck -vehicle -agasc $agasc -fid_char $fid_char -dir $mphome/$load 2>&1| tee -a $vlog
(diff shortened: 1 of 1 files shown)


jeanconn:
@taldcroft This is a monster hack, but loads the right version of starcheck.pcad_att_check, starcheck.calc_ccd_templs, and starcheck.version for every regression test run by just copying those files into the working dir.  It does this in run_regress instead of a patch for starcheck.pl.

taldcroft:
At this point what about fixing the underlying problem and have starcheck installed with pip to get a non-egg install.  I didn't realize previously that this was causing an issue.  Where the sandbox wasn't really a sandbox?

jeanconn:
Sure.  I just wanted to run the 11.10 outputs correctly before installing, and didn't want to break 11.8 starcheck just to run the test (though I suppose I could have tried to pip install 11.8, but I didn't have a clean dev ska handy to test that either way).

And yes, hadn't realized the sandbox wasn't a try starcheck Python sandbox.

jeanconn:
Also, is there a way to help pip be faster when working in a project directory with a lot of files?  At least I think that may be the problem.  In the starcheck dir it looks to just be spinning its wheels (pardon the wheel pun) and I can't tell what it is doing?

jeanconn:
Not sure if it is appropriate, but in testing I did a "python setup.py sdist" and then did a pip install of the created package.

taldcroft:
I think the best way to handle this is just put starcheck (Python) into skare at this point.  This way we will use a tested and validated install procedure going forward.

See https://github.com/sot/skare/blob/master/NOTES.fix-egg-installs for the process of installing the new de-egged package over the existing.  It is safe and reversible and can be done for a flight package without additional approvals since no code is changing.

jeanconn:
Granted, I ended up with the same procedure (make tar.gz, install it).  I was just hoping we would get to fixing up starcheck into a fully-skare-d piece before putting it into skare, instead of just having the Python part in there.

jeanconn:
Starcheck is now installed from skare, the pth file has been updated to remove any egg conflicts, and the notes have been updated in skare.  The starcheck regression tests now work against flight ska so closing this PR as not needed.

PR #152, opened 2016-06-11
Increase CCD "red" limit to -12.5C

jeanconn:
Increase CCD "red" limit 1.0C to -12.5C and yellow warning limit to -13.5 C.


taldcroft:
:+1:

PR #155, opened 2016-06-22
Add an info statement on warning free checks of big dither

jeanconn:
Add an info statement on warning free checks of big dither

If an observation with big dither passes all the checks then there is no visibility that it had big dither and that the checks were run and passed.  This PR makes sure that at least some info text is emitted.

### DIFF
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -761,6 +761,8 @@ sub large_dither_checks {
     my $all_dither = shift;
     my $time_tol = 10;         # Commands must be within $time_tol of expectation
 
+    my $warn_free = 1;
+
     # The obs tstart and tstop times are derived from the OFLS maneuver summary file,
     # which for implementation reasons gives the maneuver times offset by 10 seconds
     # from the actual command timing. Fix this here.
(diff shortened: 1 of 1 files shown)


src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -760,6 +760,8 @@ sub large_dither_checks {
     my $all_dither = shift;
     my $time_tol = 10;         # Commands must be within $time_tol of expectation
 
+    my $warn_free = 1;
```
jeanconn: I don't see a good way to do all the tests and report if any failed without some new variable.  The other option would be to have a new array to store the errors and then have the individual checks push to that local array instead of the Obsid object warn array.

taldcroft: You need a new variable, but you can do it with fewer code lines by setting a variable to the length of `@{$self->{warn}}` and then later .. `if ($#{$self->{warn}} == $n_warn) {` (or whatever the right syntax).

jeanconn: Great idea, thanks!  I can't think of a way that $self->{warn} could change length during this routine except via issuing one of these warnings, so this makes sense.

jeanconn:
I tested this directly with JAN3017A.  I haven't found a good failure case though.

jeanconn:
MAR0314A works as a "fail" test.

PR #156, opened 2016-06-29
Update text for guide star as mon with bad position

jeanconn:
Update text for guide star as mon with bad position


taldcroft:
:+1:

taldcroft:
FYI this change is also driven by requiring unique warning/error messages in review_aca_checks.

PR #160, opened 2016-07-27
Calculate the expected aca offsets 

jeanconn:
Calculate the expected aca offsets using the starcheck run of the ACA model and the aimpoint information in the zero offset table of the OR list if available.

### DIFF
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -201,7 +201,7 @@ def get_interval_data(intervals, times, ccd_temp, obsreqs=None):
         stop_idx = 1 + np.searchsorted(times, interval['tstop'])
         start_idx = -1 + np.searchsorted(times, interval['tstart'])
         ok_temps = ccd_temp[start_idx:stop_idx]
-        itimes = times[start_idx:stop_idx]
+        ok_times = times[start_idx:stop_idx]
         if len(ok_temps) > 0:
             obs['ccd_temp'] = np.max(ok_temps)
             obs['n100_warm_frac'] = dark_model.get_warm_fracs(
@@ -215,7 +215,7 @@ def get_interval_data(intervals, times, ccd_temp, obsreqs=None):
                                            obsreq['chip_id'],
                                            obsreq['chipx'],
                                            obsreq['chipy'],
-                                           time=itimes,
+                                           time=ok_times,
                                            t_ccd=ok_temps)
                 obs['aca_offset_y'] = np.mean(ddy)
                 obs['aca_offset_z'] = np.mean(ddz)
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -186,9 +186,13 @@ def get_interval_data(intervals, times, ccd_temp, obsreqs=None):
     """
     Determine the max temperature and mean offsets over each interval.
 
+    If the OR list is supplied (in the obsreqs dictionary) the ACA offsets will
+    also be calculated for each interval and included in the returned data.
+
     :param intervals: list of dictionaries describing obsid/catalog intervals
     :param times: times of the temperature samples
     :param ccd_temp: ccd temperature values
+    :param obsreqs: optional dictionary of OR list from parse_cm.or_list
 
     :returns: dictionary (keyed by obsid) of intervals with max ccd_temps
     """
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -206,6 +206,8 @@ def get_interval_data(intervals, times, ccd_temp, obsreqs=None):
             obs['ccd_temp'] = np.max(ok_temps)
             obs['n100_warm_frac'] = dark_model.get_warm_fracs(
                 100, interval['tstart'], np.max(ok_temps))
+        if obsreqs is None:
+            obsreqs = {}
         if interval['obsid'] in obsreqs and len(ok_temps) > 0:
             obsreq = obsreqs[interval['obsid']]
             if 'chip_id' in obsreq:
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -38,6 +38,8 @@ import Chandra.cmd_states as cmd_states
 from Chandra.cmd_states import get_cmd_states
 import xija
 from mica.archive.aca_dark import dark_model
+from parse_cm import read_or_list
+from chandra_aca.drift import get_aca_offsets
 
 from starcheck.version import version
 
@@ -79,6 +81,8 @@ def get_options():
                         help="xija ACA model specification file")
     parser.add_argument("--char-file",
                         help="starcheck characteristics file")
+    parser.add_argument("--orlist",
+                        help="OR list")
     parser.add_argument("--traceback",
                         default=True,
                         help='Enable tracebacks')
@@ -95,7 +99,7 @@ def get_options():
 
 def get_ccd_temps(oflsdir, outdir='out',
                   json_obsids=sys.stdin,
-                  model_spec=None, char_file=None,
+                  model_spec=None, char_file=None, orlist=None,
                   verbose=1, **kwargs):
     """
     Using the cmds and cmd_states sybase databases, available telemetry, and
@@ -169,16 +173,18 @@ def get_ccd_temps(oflsdir, outdir='out',
 
     make_check_plots(outdir, states, times,
                      ccd_temp, tstart, tstop, char=char)
-
     intervals = get_obs_intervals(sc_obsids)
-    obstemps = get_interval_temps(intervals, times, ccd_temp)
+    obsreqs = None
+    if orlist is not None:
+        obsreqs = dict([(o['obsid'], o) for o in read_or_list(orlist)])
+    obstemps = get_interval_data(intervals, times, ccd_temp, obsreqs)
     return json.dumps(obstemps, sort_keys=True, indent=4,
 	                  cls=NumpyAwareJSONEncoder)
 
 
-def get_interval_temps(intervals, times, ccd_temp):
+def get_interval_data(intervals, times, ccd_temp, obsreqs=None):
     """
-    Determine the max temperature over each interval.
+    Determine the max temperature and mean offsets over each interval.
 
     :param intervals: list of dictionaries describing obsid/catalog intervals
     :param times: times of the temperature samples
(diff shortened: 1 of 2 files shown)


jeanconn:
This should be independent of the aimpoint checking code, so I made it as a PR off master.
Writing out the data to a new JSON file is obviously optional, but the output seems like it might actually have value.

taldcroft:
The "writing to JSON" bit was the main reason I thought this should go into the flight release.  It will be quite useful!

jeanconn:
I wonder if we might want to put some times in the per-obsid file (either the obsid time ranges used for the calculations or the mean time or both).  It would make the file more useful standalone.

taldcroft:
:+1: on obsid time ranges.

jeanconn:
I forgot, I think that is actually already in there 

https://github.com/sot/starcheck/blob/cef8d8482300904b9335f968a8f852a6ae72b435/starcheck/calc_ccd_temps.py#L194

So that should be a no-op in this PR.

jeanconn:
So, per-obsid output looks like this for ORs:
```
{
    "16747": {
        "aca_offset_y": 9.6966521365499059, 
        "aca_offset_z": 1.6600082221993977, 
        "ccd_temp": -14.330923258535432, 
        "n100_warm_frac": 0.19134368387074976, 
        "obsid": 16747, 
        "tstart": 574212086.184, 
        "tstop": 574237366.184
    }, 
```
and this for ERs
```
    "51165": {
        "ccd_temp": -14.784996592120551, 
        "n100_warm_frac": 0.18381438549276496, 
        "obsid": 51165, 
        "tstart": 574886904.184, 
        "tstop": 574888616.184
    }, 
```

taldcroft:
Is there a good reason to not just dump *every* column from the dynamical offsets table into the JSON record output?

jeanconn:
OK, so a couple of things:

1) these aren't from the dynamical offset file, so I don't have all of the equivalent values (aca_ra/dec/roll)
2) I was afraid that putting out more like-named-columns would add to the confusion

If you just want to assign the values from the dynamic offset file to obsids and write that out, I think that's a different PR.

taldcroft:
OK, sounds fine.  So ignore the previous comment.

jeanconn:
Do we want this for starcheck 11.13 as well?  It will only work if we do https://github.com/sot/parse_cm/pull/3 as well.  I think we want https://github.com/sot/parse_cm/pull/2 too.

jeanconn:
So this one, had a :+1: on parts, but not a :+1: overall, and does require that parse_cm PR as mentioned in the last comment.

starcheck/calc_ccd_temps.py
```diff
@@ -195,10 +201,22 @@ def get_interval_temps(intervals, times, ccd_temp):
         stop_idx = 1 + np.searchsorted(times, interval['tstop'])
         start_idx = -1 + np.searchsorted(times, interval['tstart'])
         ok_temps = ccd_temp[start_idx:stop_idx]
+        itimes = times[start_idx:stop_idx]
         if len(ok_temps) > 0:
             obs['ccd_temp'] = np.max(ok_temps)
             obs['n100_warm_frac'] = dark_model.get_warm_fracs(
                 100, interval['tstart'], np.max(ok_temps))
+        if interval['obsid'] in obsreqs and len(ok_temps) > 0:
```
taldcroft: Need `if obsreqs is None: obsreqs = {}` so this won't raise an exception for `obsreqs = None`.

starcheck/calc_ccd_temps.py
```diff
@@ -195,10 +201,22 @@ def get_interval_temps(intervals, times, ccd_temp):
         stop_idx = 1 + np.searchsorted(times, interval['tstop'])
         start_idx = -1 + np.searchsorted(times, interval['tstart'])
         ok_temps = ccd_temp[start_idx:stop_idx]
+        itimes = times[start_idx:stop_idx]
```
taldcroft: Use consistent naming for like-defined objects, which in this case would be `ok_times`.  Or you could name both even more clearly with `interval_*`.  Usually `ok` is just reserved for a single one-off boolean mask.

taldcroft:
Finished review.

jeanconn:
Comments addressed.  Will re-review with recent schedules.

jeanconn:
I've re-reviewed with the September weeks and this looks good.

starcheck/calc_ccd_temps.py
```diff
@@ -169,16 +173,18 @@ def get_ccd_temps(oflsdir, outdir='out',
 
     make_check_plots(outdir, states, times,
                      ccd_temp, tstart, tstop, char=char)
-
     intervals = get_obs_intervals(sc_obsids)
-    obstemps = get_interval_temps(intervals, times, ccd_temp)
+    obsreqs = None
+    if orlist is not None:
+        obsreqs = dict([(o['obsid'], o) for o in read_or_list(orlist)])
```
taldcroft: Don't use `o`.  One of `obs` or `obsreq` or `or_` would work.

starcheck/calc_ccd_temps.py
```diff
@@ -169,16 +173,18 @@ def get_ccd_temps(oflsdir, outdir='out',
...
+    obstemps = get_interval_data(intervals, times, ccd_temp, obsreqs)
     return json.dumps(obstemps, sort_keys=True, indent=4,
 	                  cls=NumpyAwareJSONEncoder)
 
 
-def get_interval_temps(intervals, times, ccd_temp):
+def get_interval_data(intervals, times, ccd_temp, obsreqs=None):
```
taldcroft: Docstring needs to be updated with `obsreqs`.

starcheck/calc_ccd_temps.py
```diff
@@ -195,10 +201,24 @@ def get_interval_temps(intervals, times, ccd_temp):
...
         start_idx = -1 + np.searchsorted(times, interval['tstart'])
         ok_temps = ccd_temp[start_idx:stop_idx]
+        ok_times = times[start_idx:stop_idx]
         if len(ok_temps) > 0:
             obs['ccd_temp'] = np.max(ok_temps)
             obs['n100_warm_frac'] = dark_model.get_warm_fracs(
                 100, interval['tstart'], np.max(ok_temps))
+        if obsreqs is None:
+            obsreqs = {}
+        if interval['obsid'] in obsreqs and len(ok_temps) > 0:
```
taldcroft: You can skip lines 209 and 210 and do `if obsreqs and interval['obsid'] ..`.  Or at least move 209,210 outside the loop.

starcheck/calc_ccd_temps.py
taldcroft: Use either ternary expression or full if-then-else to set `obsreqs`.   The `x = y; if c: x = z` pattern is less obvious when reading.

taldcroft: Use dict comprehension.

taldcroft:
Would it be possible or sensible to integrate these new values into the existings `obsids.json` file, e.g. in a new top-level key `thermal`?  That way we avoid a new file.

jeanconn:
> Would it be possible or sensible to integrate these new values into the
existings obsids.json file, e.g. in a new top-level key thermal? That way
we avoid a new file.

Possible, but we'd be overwriting the 'obsids.json' file in place, because
it is written out earlier in processing as the input/reference data for the
thermal model.  I thought it would be simpler and cleaner to leave these
intervals in a separate file, but don't feel strongly.

starcheck/calc_ccd_temps.py
```diff
@@ -204,23 +204,25 @@ def get_interval_data(intervals, times, ccd_temp, obsreqs=None):
...
+        # If there are no good samples, put the times in the output dict and go to the next interval
+        if len(ok_temps) == 0:
+            obstemps[str(interval['obsid'])] = obs
+            continue
+        obs['ccd_temp'] = np.max(ok_temps)
+        obs['n100_warm_frac'] = dark_model.get_warm_fracs(
+            100, interval['tstart'], np.max(ok_temps))
+        # If we have an OR list, the obsid is in that list, and the OR list has zero-offset keys
+        if (obsreqs is not None and interval['obsid'] in obsreqs
+                and 'chip_id' in obsreqs[interval['obsid']]):
```
jeanconn: I deviated slightly from your recommendations here @taldcroft and tried to simplify the nested conditionals in this routine.

starcheck/calc_ccd_temps.py
jeanconn: Done.

starcheck/calc_ccd_temps.py
jeanconn: Done.

starcheck/calc_ccd_temps.py
jeanconn: Reorganized the whole method a bit.

starcheck/calc_ccd_temps.py
taldcroft: Well it's just that `o` is especially unusual as a variable name because of conflict with `0`.  I would not have complained about `x`.

taldcroft:
> Possible, but we'd be overwriting the 'obsids.json' file in place, because
it is written out earlier in processing as the input/reference data for the
thermal model.  I thought it would be simpler and cleaner to leave these
intervals in a separate file, but don't feel strongly.

I would prefer to not generate a new file for this small thing since the content fits in nicely with the existing file.  That will makes things simpler for downstream users of the starcheck JSON information and only add a few lines of starcheck code to do the merging.

jeanconn:
> only add a few lines of starcheck code to do the merging.

Well yeah, but they'll be Perl :-)

taldcroft:
No worries!

jeanconn:
OK, so the starcheck obsids.json file looks to just be an array at the top
level, so I'm not even sure about the best place to put this.

jeanconn:
But, since I don't think anything is reading the file, it probably doesn't matter if I change the whole structure.

taldcroft:
Not a problem, and I don't think we need to drive a change to the `obsids.json` file.  Assuming `obsid_infos` is the current list of dict and `obsid_temps` is the new temperature data as a dict, then (in Python):
```
for obsid_info in obsid_infos:
    obsid = obsid_info['obsid']
    if obsid in obsid_temps:  # just to be extra safe
        obsid_info['thermal'] = obsid_temps[obsid]
```

jeanconn:
OK,  I figured integrating the data as a new key in the list could also be a way to go.  That's also probably easier in Perl, as I can just re-JSON the extended data structure using the code that we already have that works.

jeanconn:
I'm surprised the Merge Pull Request thing is green, because I thought with the new reviews that I would be unable to merge this request until your review was updated.

taldcroft:
I might not have required changes. Anyway good to merge.

taldcroft:
Now looking at this PR on the computer, I agree that I required some changes and don't see any record of approving.  Curious.

jeanconn:
I assume that there's a github difference between requested and required changes.

jeanconn:
Ah.  I see on https://help.github.com/articles/about-required-reviews-for-pull-requests/ that I'd probably have to set starcheck master to a protected branch for github to enforce this, and even then, since I have admin privileges on starcheck, I'd probably still get a green merge button.

PR #162, opened 2016-07-28
Update regression outputs for version 11.12

jeanconn:
Update regression outputs for version 11.12

### DIFF
--- a/test_regress/test/2004/AUG0104/oflsa/starcheck.txt
+++ b/test_regress/test/2004/AUG0104/oflsa/starcheck.txt
@@ -27,6 +27,10 @@ Using simtrans file /data/mpcrit1/mplogs/2004/AUG0104/oflsa/History/SIMTRANS.txt
 Using ACABadPixel file from 2015014 Dark Cal 
 Using acquisition model for multiple star filter enabled
 
+------------  PROCESSING WARNING  -----------------
+
+WARNING: No aimpoint file matching /data/mpcrit1/mplogs/2004/AUG0104/oflsa/output/*_dynamical_offsets.txt
+
 ------------  CCD TEMPERATURE PREDICTION -----------------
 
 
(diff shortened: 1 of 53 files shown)


taldcroft:
What about a minimum date to prevent this regression?  Could be NOV0216.

jeanconn:
You mean a code change in starcheck 11.12?  I don't think it is worth it.
I don't consider these new "warnings" a regression.

taldcroft:
It is issuing a warning that is not valid, in the sense that load products prior to a certain date are not expected to have that file.  If I lost my contextual understanding of how load products have evolved and ran starcheck on old products then I would scratch my head wondering what is going on with the missing file.

Seems like code similar to https://github.com/sot/starcheck/blob/master/src/starcheck.pl#L688.

jeanconn:
Sure.  I love getting it perfect; I just thought you hated that date-based
complication, and that in this case warning is benign (the code doesn't
give any other unexpected outputs due to the missing file).

Will update and retest.
​

jeanconn:
What date would you like?

taldcroft:
What I hate more than date-based complication is seeing Processing Warnings in 53 files.  I tend to think of processing warnings as being bad.

As for date, the correct date of course is around now, but as a practical issue the test MAR0516 needs to get tested, so I would say March 1, 2016.  So we'll still get a warning when running on loads after that date before the transition, but it does not appear there are any regression test loads which are in 2016.

jeanconn:
> As for date, the correct date of course is around now, but as a practical issue the test 
> MAR0516 needs to get tested, so I would say March 1, 2016.

We don't need the date to be before MAR0516 for it to get tested.  The code will always use a dynamic aimpoint file if available, but will only warn if it is missing and the date is past the required date.

taldcroft:
Ah OK, then make it Aug. 21, 2016.  That's my guess for the first dynamic products.

jeanconn:
Changing this to a date-based requirement removes all of these diffs (as expected), so closing without merging.

PR #170, opened 2016-09-07
Collection of changes to run against SEP0516X

jeanconn:
Collection of changes to run against SEP0516X

### DIFF
--- a/starcheck/pcad_att_check.py
+++ b/starcheck/pcad_att_check.py
@@ -94,7 +94,9 @@ def make_pcad_attitude_check_report(backstop_file, or_list_file=None, mm_file=No
         lines.append('INFO: using dynamic offsets file {}'.format(dynamic_offsets_file))
         or_map = {or_['obsid']: or_ for or_ in or_list}
 
-        doffs = Table.read(dynamic_offsets_file, format='ascii.basic', guess=False)
+        dynamic_file_lines = open(dynamic_offsets_file).readlines()
+        no_end_comment = [l for l in dynamic_file_lines if not l.startswith("!")]
+        doffs = Table.read(no_end_comment, format='ascii.basic', guess=False)
         for doff in doffs:
             obsid = doff['obsid']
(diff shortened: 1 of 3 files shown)


jeanconn:
Output thus far at http://cxc.harvard.edu/mta/ASPECT/starcheck_dev/sep0516x.html

starcheck/pcad_att_check.py
```diff
@@ -94,7 +94,9 @@ def make_pcad_attitude_check_report(backstop_file, or_list_file=None, mm_file=No
         lines.append('INFO: using dynamic offsets file {}'.format(dynamic_offsets_file))
         or_map = {or_['obsid']: or_ for or_ in or_list}
 
-        doffs = Table.read(dynamic_offsets_file, format='ascii.basic', guess=False)
+        dynamic_file_lines = open(dynamic_offsets_file).readlines()
+        no_end_comment = [l for l in dynamic_file_lines if not l.startswith("!")]
```
taldcroft: Use iterators when possible, short generic variable names for one-offs, and plural names for lists:
```
# Filter out comment lines beginning with !
lines = [l for l in open(dynamic_offsets_file) if not l.startswith('!')]
doffs = Table.read(lines, ..)
```

jeanconn: Comment well taken, though I was expecting this comment strip of the
dynamic offset file to be the change least likely to be needed in testing
round two.

jeanconn: And it looks like this comment *is* going to stay, but is going to be a "#Schedule" instead of a "!Schedule".

taldcroft: But of course since it will be a normal comment then this code diff will go away.

jeanconn: And the first time round, I probably should have just customized the comment symbol list for the reader, but I didn't actually see the astropy interface for that in the docs.

PR #171, opened 2016-09-13
Warn if obsid in dynamic list and not OR

jeanconn:
Warn if obsid in dynamic list and not OR, but not the inverse.
Extra OR list entries allowed/expected.


jeanconn:
I tested this by confirming the warning is gone for 18205 in SEP0516B, and by removing an OR entry and confirming that there's a new warning like:

```
WARNING: Obsid in dynamic offsets table but missing in OR list [18125]
```

Issue #169, opened 2016-09-07
Remove red plotting of violation of planning limit


jeanconn:

jeanconn:
Closed by #173

PR #173, opened 2016-09-23
Update CCD limit to -11.5 and remove all red warnings

jeanconn:
This updates the lines in the CCD temperature plots and stops printing temperature warnings (will print an INFO statement if -11.5 is exceeded).


jeanconn:
Closes #169

jeanconn:
I should note that starcheck is also set to use the planning limit as the CCD temperature for calculating the mode for observations for which temperature could not be determined.  This change will bring some diffs in the "regress" outputs.

PR #174, opened 2016-10-07
Release 11.13 regress diffs (due to planning limit increase)

jeanconn:
Release 11.13 regress diffs (due to planning limit increase)


jeanconn:
The historical "NONE*" catalogs use the planning limit as the reference temperature for model runs, so an update to the limit causes some diffs in these outputs.

PR #179, opened 2017-01-09
Modularize Makefile to install lib, data, and bin separately

taldcroft:
This is just a lazy idea to get around being forced to install `data` that will overwrite flight data even for a test ska (assuming ska-level links of `data` and `share`).

Fully packaging the data is a better long-term idea but will be more work.


jeanconn:
I think this is no longer needed.

PR #183, opened 2017-03-16
Tweaks to support gyro bias jog / large maneuver error mitigation

jeanconn:
Tweaks to support gyro bias jog / large maneuver error mitigation

### DIFF
--- a/src/lib/Ska/Starcheck/FigureOfMerit.pm
+++ b/src/lib/Ska/Starcheck/FigureOfMerit.pm
@@ -28,8 +28,8 @@ use Inline Python => q{
 
 from chandra_aca.star_probs import acq_success_prob, prob_n_acq, mag_for_p_acq
 
-def _acq_success_prob(date, t_ccd, mag, color, spoiler):
-    out = acq_success_prob(date, float(t_ccd), float(mag), float(color), spoiler)
+def _acq_success_prob(date, t_ccd, mag, color, spoiler, halfwidth):
+    out = acq_success_prob(date, float(t_ccd), float(mag), float(color), spoiler, int(halfwidth))
     return out.tolist()
 
 def _prob_n_acq(acq_probs):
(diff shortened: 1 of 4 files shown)


src/lib/Ska/Starcheck/FigureOfMerit.pm
```diff
@@ -28,8 +28,8 @@ use Inline Python => q{
 
 from chandra_aca.star_probs import acq_success_prob, prob_n_acq, mag_for_p_acq
 
-def _acq_success_prob(date, t_ccd, mag, color, spoiler):
-    out = acq_success_prob(date, float(t_ccd), float(mag), float(color), spoiler)
+def _acq_success_prob(date, t_ccd, mag, color, spoiler, halfwidth):
+    out = acq_success_prob(date, float(t_ccd), float(mag), float(color), spoiler, int(halfwidth))
```
jeanconn: I seem to recall that the explicit casts are in here because the Inline::Python will pass the thing as a string or some such.

src/starcheck.pl
```diff
@@ -624,6 +624,7 @@ sub json_obsids{
     $obs{$obsid}->check_for_special_case_er();
     $obs{$obsid}->check_bright_perigee($radmon);
     $obs{$obsid}->count_good_stars();
+    $obs{$obsid}->check_big_box_stars();
```
jeanconn: This isn't required, but as a reviewer, I'd prefer to have this or something like it in there to tide us over.

src/lib/Ska/Starcheck/FigureOfMerit.pm
taldcroft: Were you reading my mind yesterday?

src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -2515,6 +2515,27 @@ sub count_good_stars{
 
 }
 
+###################################################################################
+sub check_big_box_stars{
```
taldcroft: I like the designation of a "big-box star".

src/lib/Ska/Starcheck/FigureOfMerit.pm
jeanconn: But apparently not enough to check this and add a helpful comment to the actual code in this PR

Issue #184, opened 2017-03-17
Feed correct spoiler status to chandra_aca.star_probs


jeanconn:
Feed correct spoiler status to chandra_aca.star_probs

jeanconn:
It looks like this line:

https://github.com/sot/starcheck/blob/9f22ad06a785c522e9aafdcebe7335bc303658a3/src/lib/Ska/Starcheck/FigureOfMerit.pm#L58

is looking for warnings about "Search Spoiler"s and unfortunately, the spoiler warnings are "Search spoiler".  Note the case on the S in spoiler.  So this should either be made into a case insensitive match or the case should be fixed.

I noticed this when looking at test products for the bigger box change:

http://cxc.cfa.harvard.edu/mta/ASPECT/starcheck_dev/starcheck_bigger_boxes/test/MAR0617A/starcheck.html#obsid20029

The bigger box should have both reduced the probability of IDX 10 because of the larger box *and* because the box now technically has a spoiler warning.  The spoiler status is being silently ignored.

jeanconn:
Close by #185

PR #185, opened 2017-03-17
Fix case on search for 'Search spoiler' in warnings

jeanconn:
Fix case on search for 'Search spoiler' in warnings.  

Incorrect case on the "s" in the regular expression to match a "Search spoiler" warning in FigureOfMerit means that the expression is never matched and all stars are assigned a spoiler status of 0.  This means that acquisition probabilities are incorrect for all stars that have spoilers (the ones already noticed by the starcheck checks for spoilers).  This PR fixes the case of the "s" in the regular expression and makes the regular expression case insensitive for good measure.  The overall impact of the change is small because SAUSAGE does not select many stars with known spoilers.

### DIFF
--- a/src/lib/Ska/Starcheck/FigureOfMerit.pm
+++ b/src/lib/Ska/Starcheck/FigureOfMerit.pm
@@ -55,7 +55,7 @@ sub make_figure_of_merit{
 	if ($c->{"TYPE$i"} =~ /BOT|ACQ/) {
 	    my $mag = $c->{"GS_MAG$i"};
             my @warnings = grep {/\[\s{0,1}$i\]/} (@{$self->{warn}}, @{$self->{yellow_warn}});
-            my $spoiler = grep(/Search spoiler/, @warnings) ? 1 : 0;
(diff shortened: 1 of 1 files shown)


jeanconn:
Closes #184

jeanconn:
@taldcroft let me know if you want this in for tomorrow too and I'll rerun everything.  I think that if we haven't been bit by this it can probably just wait one release.

taldcroft:
Maybe we can get this in next week.  It does seem like a relatively high priority.  I think you don't have to re-do all the other regression/functional testing with this in, however.

jeanconn:
Yes, I called out in my issue text that we could fix this either with the right case or a case insensitive search.  I thought the right case was just a more obvious code change, though there's nothing to prevent doing *both*.

jeanconn:
And regarding grepping our own warnings, I agree it does look odd. I suppose the problem there really is the we don't store per-star "interesting bit of info from our other checks" except as the warnings.  Could start a better data structure to do that.

jeanconn:
And yes, while relatively high priority this should still be low impact (just because SAUSAGE is pretty good about not selecting spoiled stars).

PR #187, opened 2017-03-21
Use mica acq stats database (hdf5) for acquisition histories

jeanconn:
Use mica acq stats database (hdf5) for acquisition histories.  

Flight starcheck is presently reading star histories from two tables in Sybase.  This PR would use the acquisition histories from the mica acq stats table (an hdf5 file) instead.  The advantage of this table for starcheck is that the mica acq stats table includes the tracked star positions with the one shot retroactively applied.   That content would let us determine if a star or a spoiler was tracked during the acquisition readout (something tracked within 5 arcsecs of intended position), and from that information, we can infer if the star "would have been" acquired for the image window *if* the image window had only had the currently-respected error flags applied (saturated pixel, ionizing radiation).  Use of this "would the star have been acquired" content will help to not "penalize" stars unfairly if previous acquisitions had the multiple star flag set due to bad pixels.

This would be the first use of the mica acq stats table in a flight/operational context.

In its current form, this PR will still not make the acquisition histories available on GRETA.

### DIFF
--- a/src/lib/Ska/Starcheck/FigureOfMerit.pm
+++ b/src/lib/Ska/Starcheck/FigureOfMerit.pm
@@ -59,6 +59,13 @@ sub make_figure_of_merit{
             my $color = $c->{"GS_BV$i"};
             my $hw = $c->{"HALFW$i"};
             my $star_prob = _acq_success_prob($date, $t_ccd, $mag, $color, $spoiler, $hw);
+            if (grep(/Bad Acquisition Star/i, @warnings)){
+                my @w = grep(/Bad Acquisition Star/i, @warnings);
+                my ($failed, $total) = parse_bad_acq_warning($w[0]);
+                if ($star_prob > (($total - $failed) / $total)){
+                    $star_prob = ($total - $failed) / $total;
+                }
+            }
 	    push @probs, $star_prob;
             $slot_probs{$c->{"IMNUM$i"}} = $star_prob;
             $c->{"P_ACQ$i"} = $star_prob;
--- a/src/lib/Ska/Starcheck/FigureOfMerit.pm
+++ b/src/lib/Ska/Starcheck/FigureOfMerit.pm
@@ -94,51 +94,6 @@ sub set_dynamic_mag_limits{
 }
 
 
-##*****************************************************************************************
-sub star_prob {
-##*****************************************************************************************
-    my ($acq) = @_;
-    my @warnings = @{ $acq->{warnings} };
-    my $mag = $acq->{magnitude};
-    my $warm_frac = $acq->{n100_warm_frac};
-
-    my $p_0p7color = .4294; #probability of acquiring a B-V = 0.700 star
-    my $p_1p5color = 0.9452; #probability of acquiring a B-V = 1.500 star
-    my $p_seaspo = .9241; #probability of acquiring a search spoiled star
-
-    my $mag10 = $mag - 10.0;
-    # Minimum warm fraction seen in fit data
-    my $warm_frac_min = 0.0412;
-    # Co-efficients from 2013 State of the ACA polynomial fit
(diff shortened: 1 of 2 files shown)


src/lib/Ska/Starcheck/FigureOfMerit.pm
```diff
@@ -59,6 +59,13 @@ sub make_figure_of_merit{
             my $color = $c->{"GS_BV$i"};
             my $hw = $c->{"HALFW$i"};
             my $star_prob = _acq_success_prob($date, $t_ccd, $mag, $color, $spoiler, $hw);
+            if (grep(/Bad Acquisition Star/i, @warnings)){
```
jeanconn: Here's one idea for probability overloading.  This uses the same @warnings crap, but those warnings would now be based on the mica acq stats table (though in this split configuration with guide stats still coming from Sybase, both are still needed or the code just defaults back to the rdb files).

src/lib/Ska/Starcheck/FigureOfMerit.pm
```diff
@@ -59,6 +59,13 @@ sub make_figure_of_merit{
...
+                my @w = grep(/Bad Acquisition Star/i, @warnings);
+                my ($failed, $total) = parse_bad_acq_warning($w[0]);
```
jeanconn: By using the pre-existing Bad Acquisition warnings, the probabilities will be overridden only if a warning is popping up.  The warnings right now are for bad fraction > 0.3 and more than 3 observations.

src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -2265,6 +2265,31 @@ sub identify_stars {
...
 }
 
+use Inline Python => q{
+
+from Chandra.Time import DateTime
+import numpy as np
+from astropy.table import Table
+from mica.stats.acq_stats import get_stats
+
+ACQS = get_stats()
```
jeanconn: This just reads... everything from the table.  It would probably be useful to add a routine in mica acq stats to get this by star.

jeanconn:
I'll rebase and clean this up.  What is required testing/validation for using the mica table?

src/lib/Ska/Starcheck/Obsid.pm
taldcroft: Yes, we should be able to add an index by AGASC ID and make very efficient queries.

Having said that, it isn't completely obvious that total run time will be reduced if you think about making lots of disk queries for individual stars vs one bulk disk query and then memory queries.

jeanconn:
On Apr 4, 2018 12:37 PM, "Tom Aldcroft" <notifications@github.com> wrote:

*@taldcroft* commented on this pull request.
------------------------------

In src/lib/Ska/Starcheck/Obsid.pm
<https://github.com/sot/starcheck/pull/187#discussion_r179206680>:

> @@ -2265,6 +2265,31 @@ sub identify_stars {
     }
 }

+use Inline Python => q{
+
+from Chandra.Time import DateTime
+import numpy as np
+from astropy.table import Table
+from mica.stats.acq_stats import get_stats
+
+ACQS = get_stats()

Yes, we should be able to add an index by AGASC ID and make very efficient
queries.


The index should already be there (at least for the mini agasc).

https://github.com/sot/agasc/blob/master/create_mini_agasc_h5.py#L41


Having said that, it isn't completely obvious that total run time will be
reduced if you think about making lots of disk queries for individual stars
vs one bulk disk query and then memory queries.

I was thinking that the Perl/Python interface hacks especially may make the
implementation cleaner if done with per-star queries, but will take a hard
look.

jeanconn:
When we back to this, confirm that star magnitudes for untracked / unacquired stars aren't being averaged in to the average magnitudes as 0 mags.

jeanconn:
OBE

PR #192, opened 2017-04-03
Add red warn on GUI/BOT fainter than dynamic yellow limit

jeanconn:
Add red warn on GUI/BOT fainter than dynamic yellow limit

Strictly speaking, this doesn't *add* a red warning so much as change a yellow warning to a red warning on magnitude if the IDX is for a guide or both star.  The plan here is that we want to be more conservative about picking faint stars as guide stars and would rather command no star than a faint one.


taldcroft:
This is superceded by #193, right?

jeanconn:
Yes, I just coded and tested both options, but we'll only do one.

taldcroft:
OK, since #193 is approved I'm closing this PR.

PR #190, opened 2017-03-23
Add end of maneuver date/time to output for obsid

jeanconn:
Add end of maneuver date/time to output for obsid.

This PR is a response to a feature request from Tom to add the end-of-the-maneuver time / beginning of NPM dwell time somewhere to the obsid output.  The other request was to try not to increase the overall length in lines of the obsid output, so the first draft of this PR stuck the content at the end of an existing line.   The revised output adds this information to the Maneuver section as a new "Ends=" key/value:
```
MP_TARGQUAT at 2013:302:14:59:33.091 (VCDU count = 6607454)
  Q1,Q2,Q3,Q4: -0.16732482  0.81354749  0.37632263  0.41051696
  MANVR: Angle= 103.97 deg  Duration= 2046 sec  Slew err= 74.3
  MANVR: Ends= 2013:302:15:33:34.000
```

PR #196, opened 2017-04-05
Add end of maneuver date/time to output for obsid

jeanconn:
Add end of maneuver date/time to output for obsid

These are the commits that were approved and merged as #190 , but it looks like *this* is exactly where my accidental force push to master overwrote the record of that merge ever happening.  Redoing the PR seems the best way to proceed.


jeanconn:
I'm not sure if I restored the previous branch if it would let me "re"merge, but for 4 commits it doesn't seem critical to experiment with github that way.

taldcroft:
Do we want to disallow force push to master for starcheck?

jeanconn:
Yes.

jeanconn:
​I just set "master" as a protected branch and also checked the box to
apply that to adminstrators.​

PR #197, opened 2017-04-13
Update ACA model to correspond to chandra_models 0.12 version

jeanconn:
Update ACA model to correspond to chandra_models 0.12 version


jeanconn:
@taldcroft Am I cleared to just cut this as a starcheck release for this week?  I figure smaller releases are probably better anyway.

PR #199, opened 2017-04-28
Add bad pixel check for ACQ box

jeanconn:
Add bad pixel check for ACQ box and set bad pixel check on guide stars to work for those selected at any stage (looks like previous code was only stage 1 or stage2)

### DIFF
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -1374,22 +1374,37 @@ sub check_star_catalog {
 	# Bad Pixels ACA-025
         my @close_pixels;
         my @dr;
-	if ($type ne 'ACQ' and $c->{"GS_PASS$i"} =~ /^1|\s+|g[1-2]/) {
+        my @dz;
+        my @dy;
+	if ($type ne 'MON' and $type ne 'FID'){
 	    foreach my $pixel (@bad_pixels) {
 		my $dy = abs($yag-$pixel->{yag});
 		my $dz = abs($zag-$pixel->{zag});
 		my $dr = sqrt($dy**2 + $dz**2);
-		next unless ( $dz < $dither+25 and $dy < $dither+25 );
+		next unless ( $dz < $c->{"HALFW$i"} and $dy < $c->{"HALFW$i"});
 		push @close_pixels, sprintf("%3d, %3d, %3d\n", $dy, $dz, $dr);
+                push @dy, $dy;
+                push @dz, $dz;
 		push @dr, $dr;
-	    }   
+	    }
 	    if ( @close_pixels > 0 ) {
 		my ($closest) = sort { $dr[$a] <=> $dr[$b] } (0 .. $#dr);
-		my $warn = sprintf("$alarm [%2d] Nearby ACA bad pixel. " .
-				   "Y,Z,Radial seps: " . $close_pixels[$closest],
-				   $i); #Only warn for the closest pixel
-		push @warn, $warn;
-	    }
+                if ($type eq 'BOT' or $type eq 'GUI'){
+                    if ($dy[$closest] < $dither + 25 and $dz[$closest] < $dither + 25){
+                        push @warn, sprintf("$alarm [%2d] Nearby ACA bad pixel. " .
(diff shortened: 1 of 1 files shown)


jeanconn:
Maybe these should be informational-only warnings?


new_warning_free

example_obsid

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1374,22 +1374,37 @@ sub check_star_catalog {
...
-		next unless ( $dz < $dither+25 and $dy < $dither+25 );
+		next unless ( $dz < $c->{"HALFW$i"} and $dy < $c->{"HALFW$i"});
 		push @close_pixels, sprintf("%3d, %3d, %3d\n", $dy, $dz, $dr);
+                push @dy, $dy;
+                push @dz, $dz;
 		push @dr, $dr;
-	    }   
+	    }
 	    if ( @close_pixels > 0 ) {
 		my ($closest) = sort { $dr[$a] <=> $dr[$b] } (0 .. $#dr);
```
jeanconn: This still just looks for the closest pixel (no change from previous behavior).  Do we care if an acq box has more than one bad pixel?

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1374,22 +1375,37 @@ sub check_star_catalog {
...
-	if ($type ne 'ACQ' and $c->{"GS_PASS$i"} =~ /^1|\s+|g[1-2]/) {
+        my @dz;
+        my @dy;
+	if ($type ne 'MON' and $type ne 'FID'){
 	    foreach my $pixel (@bad_pixels) {
 		my $dy = abs($yag-$pixel->{yag});
 		my $dz = abs($zag-$pixel->{zag});
 		my $dr = sqrt($dy**2 + $dz**2);
-		next unless ( $dz < $dither+25 and $dy < $dither+25 );
+		next unless ( $dz < $c->{"HALFW$i"} and $dy < $c->{"HALFW$i"});
```
taldcroft: It seems this is reducing the box check size by `$dither` for guide stars.  Should this expression just be adding `$dither`?

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1374,22 +1375,37 @@ sub check_star_catalog {
...
-	    }
+                if ($type eq 'BOT' or $type eq 'GUI'){
+                    if ($dy[$closest] < $dither + 25 and $dz[$closest] < $dither + 25){
+                        push @warn, sprintf("$alarm [%2d] Nearby ACA bad pixel. " .
+                                                "Y,Z,Radial seps: " . $close_pixels[$closest],
+                                            $i); #Only warn for the closest pixel
+                    }
+
+                }
+                # BOT stars will get two warnings if within dither
```
taldcroft: I suspect that we will get a lot of these info messages and just ignore them.  Based on ASVT testing they don't significantly impact star acq probs.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1374,22 +1375,37 @@ sub check_star_catalog {
 	# Bad Pixels ACA-025
         my @close_pixels;
         my @dr;
-	if ($type ne 'ACQ' and $c->{"GS_PASS$i"} =~ /^1|\s+|g[1-2]/) {
```
taldcroft: Agree to not filter on stage.

jeanconn:
My impression was that once you looked at the scope of the warnings, we decided not to add the warning for acq pixel in search box.  If we're planning to get rid of bad pixels in acq selection, this also becomes moot.  So this PR should just be revised to just be the guide stage fix.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: The intent here was to use the bigger acquisition box to save more bad pixels for checking in the next conditional.   You are absolutely correct that this probably backfires for GUI stars which do have HALFW defined (and it is going to be smaller than dither + 25).  I think the change at this line is not necessary at all if we are getting rid of the checking for bad pixels in acq boxes.

PR #206, opened 2017-07-10
Update version to 11.18

jeanconn:
Update starcheck to version 11.18

This release is only to include use of the updated chandra_aca.plot routine which now works in row/col instead of yag/zag.  The plots are still labeled in yag/zag.


jeanconn:
This probably doesn't need a PR, but I figured this gave me a place to put one more old vs new plot:

old:
old_stars_17117

new:
new_stars_17117

The position of star 502402704 relative to the the CCD edge in the lower right quadrant was one of the reasons that there was a driver to update to a more accurate representation of star positions.

jeanconn:
Approved at today's load review; merging now and will install after products are officially approved.

PR #207, opened 2017-07-11
Use bad pixel list from 2017185 calibration process

jeanconn:
Use bad pixel list from 2017185 calibration process


jeanconn:
My thought was to put this in starcheck 11.19 .

jeanconn:
Approved at load review on 1-Aug-2017.

PR #214, opened 2017-11-27
Change logic warning for too few guide stars.

jeanconn:
Print a red warning for fewer than 4 guide stars on a science observation.
Print a yellow warning for fewer than 5 guide stars on a science observation.
The line that changed the threshold for the number of stars in the presence
of a monitor window has been cut.

### DIFF
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -1072,7 +1072,7 @@ sub check_star_catalog {
    
     my $is_science = ($self->{obsid} =~ /^\d+$/ && $self->{obsid} < $ER_MIN_OBSID);
     my $is_er      = ($self->{obsid} =~ /^\d+$/ && $self->{obsid} >= $ER_MIN_OBSID);
-    my $min_guide    = $is_science ? 5 : 6; # Minimum number of each object type
+    my $min_guide    = $is_science ? 4 : 6; # Minimum number of each object type
     my $min_acq      = $is_science ? 4 : 5;
     my $min_fid      = 3;
     ########################################################################
@@ -1124,8 +1124,6 @@ sub check_star_catalog {
 	push @{$self->{warn}}, "$alarm No star catalog for obsid $obsid ($oflsid). \n";		    
 	return;
     }
-    # Reset the minimum number of guide stars if a monitor window is commanded
-    $min_guide -= scalar grep { $c->{"TYPE$_"} eq 'MON' } (1..16);
 
     print STDERR "Checking star catalog for obsid $self->{obsid}\n";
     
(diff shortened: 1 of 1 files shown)


jeanconn:
This is just an idea for a way we could go with this.  I note that

1) This yellow-warns on all monitor windows because I've removed that logic.  As a reviewer, I think this is actually the right thing at this point.  I think monitor window configurations deserve an extra look to see if we can really spare the star.

2) For observations with 3 or fewer guide stars, this logic will print the warning about having too few guide stars twice, once in red and once in yellow.  I think this is acceptable because if/else ing the block would add code complexity without much return.

jeanconn:
Obsid 20124 example, original / as-run output (ignoring the fact that I apparently highlighted Too Few before screen capping).

orig_too_few

Output with this PR
new_too_few

taldcroft:
:+1: on the idea, but we need to present this to SSAWG.

I don't think the code complexity of the `if / elsif` is too great.  Keeping the output free of redundant warnings is worth it.

taldcroft:
In a related request, could you throw in an informational message at the end of monitor window checking that says roughly "Monitor window special commanding meets requirements"?

jeanconn:
I left this section with the postfix conditionals for these warnings, but refined the conditions for green/yellow warning a bit so it should not show up for science observations if there is already a red warning.

taldcroft:
The postfix conditionals are logically correct but hard to read because you need to make the inference that `$n_gui >= $min_guide` is the opposite of the previous condition.  How about just making it simple if / elsif:
```
if ($n_gui < $min_guide) {
    # Red warn for fewer than the minimum required guide stars
    push @warn,"$alarm Only $n_gui Guide Stars \n" ;
elsif ($is_science && $n_gui < 5) {
    # Yellow warn for fewer than nominal number of guide stars in science observations
    push @yellow_warn, "$alarm Only $n_gui Guide Stars \n";
}
```

jeanconn:
I think I mentioned it at SSAWG but didn't actually present or email to the group.  Will do.

jeanconn:
It sounds like this should be reworked a bit based on the SSAWG today.

My impression is we want to go back to a red warning for fewer than 5 stars in a science observation unless that observation has a Monitor window.

The question I had in the meeting is if we need to have a new check on the quality of the 4 stars in a monitor window observation before that should get a warning.   Probably not for now as long as Monitor windows are highlighted in the summary.

taldcroft:
It seems like the primary action here after the meeting today is just to make no change to existing code, i.e. close this PR.  The presence of a MON is already highlighted in the detail section, so that is good enough for me.  But if you or Eric want something highlighting the presence of a MON window in the summary then that's OK.

jeanconn:
I think basically, the primary driver of this PR was to update "Too few guide stars" to "Only N Guide stars" so the 3 guide star case would have been caught sooner.  

I might as well just cut this PR down to that change.

jeanconn:
Or, as you say, just close.  Will open new PR with just that change.

PR #215, opened 2017-11-27
Reorganize dither checking (mostly for ERs)

### LINKED ISSUE #247: Downgrade bad pixel warning from red to orange in some cases
Consider demoting Nearby ACA bad pixel warnings from Red to Orange for cases that are currently Red but have max(Y,Z sep) > dither + 20 + 10 (arcsec).  That keeps the pixel out of the readout window with two pixels of margin.  

Definitely with 2 pixels of margin I would rarely call for a star catalog edit.  Exactly one pixel is a little tight and could result in a hot pixel getting into the background pixels for a credible attitude disturbance.

jeanconn:
We changed to using only commanded dither (from backstop) to check dither and catalogs for observations.  However, the code that was supposed to *set* the values of the commanded dither for each obsid object was not actually run for ERs.  Basically, setting the values was a side effect of dither checking code that wasn't run on ERs.

This PR both lets that dither checking code do some checks that could actually apply to ERs and lets that "side effect" work, appropriately setting the commanded dither for ERs.

Before this PR (but after the use-only-commanded-dither change), ERs were just being checked with default 20 arcsec dither.  They are still checked with 20 arcsec dither if placed in the schedule before any dither commanding (no commanded dither available).

This PR also prints dither info for ERs in the starcheck report and adds a comment when the dither values being used for checking are just the 20 arcsec amplitude default values.

Closes #247.

### DIFF
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -1807,9 +1807,11 @@ sub print_report {
     if ( defined $self->{cmd_dither_status} ) {
 	$o .= sprintf "Dither: %-3s ",$self->{cmd_dither_status};
 	$o .= sprintf ("Y_amp=%4.1f  Z_amp=%4.1f  Y_period=%6.1f  Z_period=%6.1f",
-		       $self->{cmd_dither_y_amp}*3600., $self->{cmd_dither_z_amp}*3600.,
+		       $self->{cmd_dither_y_amp}, $self->{cmd_dither_z_amp},
 		       $self->{cmd_dither_y_period}, $self->{cmd_dither_z_period})
         if ($self->{cmd_dither_status} eq 'ENAB' && $self->{cmd_dither_y_period} && $self->{cmd_dither_z_period});
+	$o .= sprintf ("(no commanded dither amp: checked with 20 arcsec dither)")
+        if ($self->{cmd_dither_status} eq 'ENAB' && not(defined $self->{cmd_dither_y_period}));
 	$o .= "\n";
     }
 
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -622,9 +622,6 @@ sub check_dither {
     my $obs_end_pad = 3*60;
     my $manvr;
 
-    if ( $self->{obsid} =~ /^\d*$/){
-	return if ($self->{obsid} >= $ER_MIN_OBSID); # For eng obs, don't have OR to specify dither
-    }
     # If there's no starcat on purpose, return
     if (defined $self->{ok_no_starcat} and $self->{ok_no_starcat}){
         return;
@@ -654,8 +651,59 @@ sub check_dither {
         }
     }
 
-    my $bs_val = $dither->{state};
+    $self->{cmd_dither_status} = $dither->{state};
+    $self->{cmd_dither_y_amp} = $dither->{ampl_y};
+    $self->{cmd_dither_z_amp} = $dither->{ampl_p};
+    $self->{cmd_dither_y_period} = $dither->{period_y};
+    $self->{cmd_dither_z_period} = $dither->{period_p};
+
+
+    # Check for standard and large dither based solely on backstop/history values
+    # This needs to be before the checking for dither during the observation, because the pad
+    # for that checking is set here
+    if (($dither->{state} eq 'ENAB') and (defined $dither->{ampl_y} and defined $dither->{ampl_p})){
+        if (not standard_dither($dither)){
+            push @{$self->{yellow_warn}}, "$alarm Non-standard dither\n";
+            if ($dither->{ampl_y} > $large_dith_thresh or $dither->{ampl_p} > $large_dith_thresh){
+                $self->large_dither_checks($dither, $dthr);
+                # If this is a large dither, set a larger pad at the end, as we expect
+                # standard dither parameters to be commanded at 5 minutes before end,
+                # which is greater than the 3 minutes used in the "no dither changes
+                # during observation check below
+                $obs_end_pad = 5.5 * 60;
+            }
+        }
+    }
+
+
+    # Loop to check for dither changes during the observation
+    # ACA-003
+    if (not defined $obs_tstop ){
+        push @{$self->{warn}},
+            "$alarm Unable to determine obs tstop; could not check for dither changes during obs\n";
+    }
+    else{
+        foreach my $dither (reverse @{$dthr}) {
+            if ($dither->{time} > ($obs_tstart + $obs_beg_pad)
+                    && $dither->{time} <= $obs_tstop - $obs_end_pad) {
+                push @{$self->{warn}}, "$alarm Dither commanding at $dither->{time}.  During observation.\n";
+            }
+            if ($dither->{time} < $obs_tstart){
+                last;
+            }
+        }
+    }
+
+
+
+
+    # For eng obs, don't have OR to specify dither, so stop before doing vs-OR comparisons
+    if ( $self->{obsid} =~ /^\d*$/){
+	return if ($self->{obsid} >= $ER_MIN_OBSID);
+    }
+
     # Get the OR value of dither and compare if available
+    my $bs_val = $dither->{state};
     my $or_val;
     if (defined $self->{DITHER_ON}){
         $or_val = $dthr_cmd{$self->{DITHER_ON}};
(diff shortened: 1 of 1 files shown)


jeanconn:
WIP because I don't recall if these changes already do the checking for bad pixels and such using the new dither for ERs.

jeanconn:
Not sure about my WIP comment above, but the reorganization of the check_dither routine now does set the cmd_dither parameters correctly for ERs (if there has been dither commanding at some point earlier in backstop).  The cmd_dither parameters are used, if set, to do the catalog checking including the bad pixel check.  If cmd_dither parameters not set, they default to 20 arcsec amp as before.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -622,9 +622,6 @@ sub check_dither {
     my $obs_end_pad = 3*60;
     my $manvr;
 
-    if ( $self->{obsid} =~ /^\d*$/){
-	return if ($self->{obsid} >= $ER_MIN_OBSID); # For eng obs, don't have OR to specify dither
```
jeanconn: This is basically just moved later in the method so that the cmd_dither_* values are set even for ERs.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -654,39 +651,17 @@ sub check_dither {
...
-    }
-    else{
-        push @{$self->{warn}},
-            "$alarm Unable to determine dither from OR list\n";
-    }
+    $self->{cmd_dither_status} = $dither->{state};
+    $self->{cmd_dither_y_amp} = $dither->{ampl_y};
+    $self->{cmd_dither_z_amp} = $dither->{ampl_p};
+    $self->{cmd_dither_y_period} = $dither->{period_y};
+    $self->{cmd_dither_z_period} = $dither->{period_p};
```
jeanconn: And this is where those cmd_dither_* values are set.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -654,39 +651,17 @@ sub check_dither {
...
-                               $dither->{ampl_y}, $dither->{ampl_p});
-            push @{$self->{warn}}, $warn;
-        }
-    }
     # Check for standard and large dither based solely on backstop/history values
-    if (($bs_val eq 'ENAB') and (defined $dither->{ampl_y} and defined $dither->{ampl_p})){
-        $self->{cmd_dither_y_amp} = $dither->{ampl_y};
-        $self->{cmd_dither_z_amp} = $dither->{ampl_p};
+    # This needs to be before the checking for dither during the observation, because the pad
+    # for that checking is set here
```
jeanconn: Just a comment update here.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1790,12 +1804,14 @@ sub print_report {
     if ( ( defined $self->{ra} ) and (defined $self->{dec}) and (defined $self->{roll})){
 	$o .= sprintf "RA, Dec, Roll (deg): %12.6f %12.6f %12.6f\n", $self->{ra}, $self->{dec}, $self->{roll};
     }
-    if ( defined $self->{DITHER_ON} && $self->{obsid} < $ER_MIN_OBSID ) {
-	$o .= sprintf "Dither: %-3s ",$self->{DITHER_ON};
+    if ( defined $self->{cmd_dither_status} ) {
+	$o .= sprintf "Dither: %-3s ",$self->{cmd_dither_status};
```
jeanconn: This change changes the text from 'ON' to ENAB, so that's somewhat controversial.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1790,12 +1804,14 @@ sub print_report {
...
 	$o .= sprintf ("Y_amp=%4.1f  Z_amp=%4.1f  Y_period=%6.1f  Z_period=%6.1f",
-		       $self->{DITHER_Y_AMP}*3600., $self->{DITHER_Z_AMP}*3600.,
-		       360./$self->{DITHER_Y_FREQ}, 360./$self->{DITHER_Z_FREQ})
-	    if ($self->{DITHER_ON} eq 'ON' && $self->{DITHER_Y_FREQ} && $self->{DITHER_Z_FREQ});
+		       $self->{cmd_dither_y_amp}, $self->{cmd_dither_z_amp},
+		       $self->{cmd_dither_y_period}, $self->{cmd_dither_z_period})
+        if ($self->{cmd_dither_status} eq 'ENAB' && $self->{cmd_dither_y_period} && $self->{cmd_dither_z_period});
+	$o .= sprintf ("(no commanded dither amp: checked with 20 arcsec dither)")
```
jeanconn: And this adds a new comment that mentions the amplitude used for the actual catalog check.  It just sticks it on the line previously reserved for the actual dither parameters.  This is also controversial.  And will probably need accompanying change in starcheck_parser code.

jeanconn: Are we ready to use something for continuity for dither params?

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -717,6 +693,44 @@ sub check_dither {
...
+    # For eng obs, don't have OR to specify dither, so stop before doing vs-OR comparisons
+    if ( $self->{obsid} =~ /^\d*$/){
+	return if ($self->{obsid} >= $ER_MIN_OBSID);
+    }
+
+    # Get the OR value of dither and compare if available
+    my $bs_val = $dither->{state};
+    my $or_val;
+    if (defined $self->{DITHER_ON}){
+        $or_val = $dthr_cmd{$self->{DITHER_ON}};
```
taldcroft: This redirection is just confusing and makes the code harder to read.  At least make the variable name descriptive like `$translate_ON_OFF_to_ENAB_DISA`.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -717,6 +693,44 @@ sub check_dither {
...
+
+
+
+    # For eng obs, don't have OR to specify dither, so stop before doing vs-OR comparisons
+    if ( $self->{obsid} =~ /^\d*$/){
+	return if ($self->{obsid} >= $ER_MIN_OBSID);
+    }
+
+    # Get the OR value of dither and compare if available
+    my $bs_val = $dither->{state};
```
taldcroft: It seems like there is inconsistent use of `$dither` and `$self->{cmd_*}`.  If possible, up top you should define those `$self->{cmd_*}` values and then forget about `$dither`.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -717,6 +693,44 @@ sub check_dither {
...
+    my $or_val;
+    if (defined $self->{DITHER_ON}){
```
taldcroft: I know it's legacy, but looking at this code for the first time in forever, it is incredibly confusing that there are OR-derived values in `$self` with all caps and inconsistent names.  Shouldn't those be captured in a substructure like `$self->{OR}->{DITHER_ON}` or some-such?

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -717,6 +693,44 @@ sub check_dither {
...
+        $or_val = $dthr_cmd{$self->{DITHER_ON}};
+        # ACA-002
+        push @{$self->{warn}}, "$alarm Dither mismatch - OR: $or_val != Backstop: $bs_val\n"
```
taldcroft: When getting into a section of code, you should start making a habit of replacing these post-fix conditionals with `if (test) {do something}`.  Especially for a long one like this it is much more difficult to read because the conditional is almost hidden instead of being up-front like the way I think.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
taldcroft: I think `ENAB` is more correct, but if there are downstream interface impacts (annie etc?) then it might not be worth it.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
taldcroft: Kadi.commands is probably not ready for operational use like this.  It is close but we need to get daily checking in place and bring to LR.  So open a new issue for that.

taldcroft: I'll have to see how the output looks with the new comment.

jeanconn:
I'm thinking that this PR can be deferred (worst case is continuing to check ERs with 20 arcsec dither when they may have 8 arcsec dither) and the other PRs can go into a starcheck release for this week.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: I can open an issue and work at this if you really want to focus on cleaning up these data structures.  I don't think this is within scope on this PR.  The problem with this idea of moving them is that I don't even know immediately which ones are the OR-derived values as they just basically get bulk copied

https://github.com/sot/starcheck/blob/229b1f82f4a0a21124d0d66cff1fc82a6ae4fd81/starcheck/src/starcheck.pl#L466

Moving them into an OR substructure would be easy, but then going through every starcheck routine to redirect the references seems like it could end up large-scope.

jeanconn:
OBE #266

PR #216, opened 2017-11-29
Add info confirmation an observation passes monitor win checks

jeanconn:
Add an fyi/info statement confirming an observation passes monitor window checks.

### DIFF
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -1499,6 +1499,8 @@ sub check_monitor_commanding {
 
     my $r2a = 180./3.14159265*3600;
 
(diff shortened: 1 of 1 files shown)


jeanconn:
This seemed similar to the info statement we have that an observation passes big dither checks (which are also a subset of the monitor window checks) so I used the "count the warnings" strategy @taldcroft suggested when we put that in place.

jeanconn:
Here's an example of new output for the info statement:

new_mon_ok

jeanconn:
Here is an example of not getting the info statement:
bad_mon

I note that we also won't see the info statement if the monitor window position is off, which isn't exactly a "special commanding" issue, so we may want to tweak the comment.  My inclination was just "Observation passes monitor window checks" which is less specific but perhaps more accurate.

Issue #209, opened 2017-08-09
Account for zodiacal brightness in estimating star probability


taldcroft:
The dark current modeling removes zodiacal brightness, which in turn underestimates the warm fraction in general.  For the most accurate results we should included the estimated zodiacal brightness for each field.

jeanconn:
This gives a driver to Python-ize the simple zodiacal brightness lookup and put it somewhere.  Should that live in starcheck, a new module, chandra_aca, or Ska.astro, or?

taldcroft:
Probably `chandra_aca.star_probs` would work.  BTW, one short-term workaround is just to increase the warm_frac by a fixed value that accounts for the mean zodi brightness.

jeanconn:
Right, I suppose the short term work-around makes sense.  Right now I don't think we have anything in Python to convert to ecliptic coordinates (needed for the table lookup).  I haven't looked at the release notes to know if we'd be fine updating astropy to a version >= 1.1 in PY2 HEAD flight for use with this.

jeanconn:
Do you have a "fit" value for mean zodi brightness that I should use?  I assume that we want at least the short-term workaround for starcheck 11.21?

jeanconn:
Regarding actually implementing this, once I have the zodiacal brightness in e-/sec for a field, do we need an offset option for `chandra_aca.dark_model.get_warm_fracs` or would it be obvious enough if the calling code in `chandra_aca.star_probs` just adjusted the warm threshold due to the zodiacal brightness contribution (as in 5 e-/sec zodiacal brightness moves N100 warm threshold to 95).

taldcroft:
Coming back to this after seeing the implementation, I think I disagree with myself circa Aug 2017.  
```
warm_frac = get_warm_fracs(WARM_THRESHOLD - bgd, date=date, T_ccd=t_ccd)
```
This has an obvious problem that if `bgd` is more than `WARM_THRESHOLD` then everything blows up, even though in reality we could acquire stars with a uniform background of 100 e-/sec.

There is a deeper issue, which is that the acq probability model is *calibrated* using the warm frac derived from the zodi-subtraced dark cal.  So warm frac is an indicator of dark current *non-uniformity*, which correlates with star acquisition.  But we do not actually know how star acquisition probs are impacted by adding a uniform bias (zodi bgd).  Here we could use ASVT to potentially add bgd level as another model parameter.

[...]

PR #217, opened 2017-12-01
Add a path to calculate and use zodiacal brightness estimate for acq probs

jeanconn:
Add a path to calculate and use zodiacal brightness estimate for acq probs.  Per #209, "The dark current modeling removes zodiacal brightness, which in turn underestimates the warm fraction in general", so this PR would add a new calculation of zodiacal brightness for each attitude and pass that estimate to the acquisition probability code.


starcheck/zodiacal_brightness.py
```diff
@@ -0,0 +1,58 @@
+import os
+import numpy as np
+from astropy.table import Table
+from Ska.Sun import position
+
+TABLE = os.path.join(os.path.dirname(__file__), 'data', 'table_17.csv')
+RESPONSIVITY = 0.0108 # (e-/sec/SI unit zodiacal flux)
+
+def equat2eclip(ra, dec):
```
jeanconn: I would have preferred to use ecliptic coordinates from a more modern version of astropy (support introduced in 1.1 I think), but more modern versions of astropy require updated numpy and that seemed bigger than we needed right now in Py2 Ska.  If we swap over to Py3 Ska for starcheck relatively quickly, this routine could just be cut.

Tested against the Perl port of the IDL we were already using, but I'll include some more notes about that in this PR.

jeanconn:
Still needs more test notes for equat2eclip

taldcroft:
Overtaken by proseco and the more recent acq probability modeling formalism.

PR #221, opened 2018-01-15
WIP: Get some of the dark cal reporting on fewer than 5 replicas

jeanconn:


taldcroft:
Is this at all related to SRDC at this point?

jeanconn:
I don't think so.

taldcroft:
OBE, we do not do multi-replica dark cals any more.

PR #224, opened 2018-03-09
Do bad pixel checks for all guide stars

jeanconn:
This changes the bad pixel check from working only on first and second stage guide stars to all guide stars.


starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1374,7 +1374,7 @@ sub check_star_catalog {
 	# Bad Pixels ACA-025
         my @close_pixels;
         my @dr;
-	if ($type ne 'ACQ' and $c->{"GS_PASS$i"} =~ /^1|\s+|g[1-2]/) {
+	if ($type ne 'ACQ'){
```
jeanconn: We probably just need this on GUI and BOT (not FID/MON/ACQ).

PR #225, opened 2018-03-09
Red warn on too many monitor windows

jeanconn:
Red warn on too many monitor windows ( more than 1 for OR, more than 2 for ER).
This also stops decrementing the minimum required number of commanded guide stars by the count of monitor windows if the observation is an ER.


jeanconn:
I tested this code by setting the thresholds lower temporarily.  I forgot to save those outputs to just let me know if I should redo and post in this PR.

PR #234, opened 2018-04-04
Remove quadrant boundary checks

jeanconn:
Remove quadrant boundary checks (fid and guide star) in the Obsid code.

### DIFF
--- a/starcheck/src/aca_load_review_cl.rst
+++ b/starcheck/src/aca_load_review_cl.rst
@@ -241,10 +241,6 @@ Checks
 |ACA-011|Magnitude limit   |X|X|FL: 6.8 - 7.2                              |n/a |Reduced aspect  |
 |       |                  | | |                                           |    |quality         |
 +-------+------------------+-+-+-------------------------------------------+----+----------------+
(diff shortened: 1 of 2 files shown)


jeanconn:
Also need to modify checklist

jeanconn:
I've cut ACA-013 and ACA-014.  I don't remember why ACA-012 is there (it says it is n/a?).  That can probably get cut too.

jeanconn:
@taldcroft Should I cut ACA-012?

taldcroft:
No.

jeanconn:
OK.  Do you know what it is or how to check for it?

jeanconn:
I also was just assuming that concerns about CTI were the only reason we didn't want to place fid lights on quadrant boundaries.  If there is some other calibration concern I can put back that code.

taldcroft:
The first time I looked ACA-012 I assumed the `n/a` meant we weren't doing anything with this and no action was required.  But looking again I see it should just be deleted.

jeanconn:
Thanks, yeah.  It is true that the `n/a` means we aren't doing anything with it, but it seemed weird for an `n/a` item to stay in the list if we're cutting others that are no longer applicable.

Issue #226, opened 2018-03-09
What should we do for color == 1.5 stars?


jeanconn:
Right now they get a yellow warning only if also fainter than 10.0 mag.
Should the mag condition be removed? 
Should this be a yellow or orange warning?
Only guide stars get warnings based on color in the current code.

jeanconn:
This is still an outstanding issue.  @taldcroft What are your thoughts on when you should get a warning and what printed color should that warning have for a guide star with color == 1.5?

taldcroft:
Yellow warning independent of mag.  The mag_err is appropriately large for these, and acq probability appropriately adjusted, so yellow is enough.  But as in the case shown in #232, having *no* indication of this is definitely confusing.

jeanconn:
I should expand this discussion to ACQ.  The section in 

https://github.com/sot/starcheck/blob/f6520a6f51f349c56ea8a08d3f0100218865833a/starcheck/src/lib/Ska/Starcheck/Obsid.pm#L1230

is where all this confusion happens.  Right now stars are assigned a "marginal note" if color == 1.5 *and* mag > 10.0.

If guide stars, they get an orange warning for a color=0.700 "marginal note" and a yellow warning for a 1.5 "marginal note".  

If a star is bad class it will get a warning for both acq and guide.

jeanconn:
OK. So I'll cut the mag dependent logic and let the warning print for both acq and guide stars.

jeanconn:
While I'm at it, should ACQ-only stars get an orange warning for color=0.700 ?

taldcroft:
> While I'm at it, should ACQ-only stars get an orange warning for color=0.700 ?

Sure.

jeanconn:
It looks like removing the mag filter on the 1.5 warning contributes to a lot more text warnings than I would have expected (just didn't realize we use so many of these stars).  Not sure if we want to select these later in selection as well.

jeanconn:
Closed by #233

jeanconn:
Is this noisy enough to hold promotion? (was kinda hoping to get starcheck 11.24 out today, but fine if we defer another week for some more improvements)

Do you want to stop the 1.5 green warnings altogether or make them magnitude dependent again?  I'm guessing from your text that you aren't interested in a more complicated combination of something like "if we've got a measurement of this star within N * mag_aca_err, don't bother with the color=1.5 warning).

[...]

PR #237, opened 2018-04-14
Remove all the text 1.5 color warnings and add color and mag_aca_err hover overs

jeanconn:
Remove all the text 1.5 color warnings and add color and mag_aca_err hover overs

### DIFF
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -1898,44 +1898,40 @@ sub print_report {
 		$idpad_n --;
 	    }
             my $acq_prob = "";
+            my $history_blurb = "";
             if ($c->{"TYPE$i"} =~ /BOT|ACQ/){
                 my $prob = $self->{acq_probs}->{$c->{"IMNUM${i}"}};
                 $acq_prob = sprintf("Prob Acq Success %5.3f", $prob);
+
             }
 	    if ($db_stats->{acq} or $db_stats->{gui}){
-	    $table .= sprintf("${idpad}<A HREF=\"%s%s\" STYLE=\"text-decoration: none;\" "
-			      . "ONMOUSEOVER=\"return overlib ('"
-			      . "ACQ total:%d noid:%d <BR />"
-			      . "GUI total:%d bad:%d fail:%d obc_bad:%d <BR />"
-			      . "Avg Mag %4.2f <BR />"
-                              . "$acq_prob"
-			      . "', WIDTH, 220);\" ONMOUSEOUT=\"return nd();\">%s</A>", 
-			      $acq_stat_lookup, $c->{"GS_ID${i}"}, 
-			      $db_stats->{acq}, $db_stats->{acq_noid},
-			      $db_stats->{gui}, $db_stats->{gui_bad}, $db_stats->{gui_fail}, $db_stats->{gui_obc_bad},
-			      $db_stats->{avg_mag},
+                $history_blurb = sprintf("ACQ total:%d noid:%d <BR />"
+			       . "GUI total:%d bad:%d fail:%d obc_bad:%d <BR />"
+			       . "Avg Mag %4.2f <BR />",
+                                         $db_stats->{acq}, $db_stats->{acq_noid},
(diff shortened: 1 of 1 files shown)


jeanconn:
Should finish up #226

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1894,44 +1898,40 @@ sub print_report {
...
-                              . "$acq_prob"
-			      . "', WIDTH, 220);\" ONMOUSEOUT=\"return nd();\">%s</A>", 
-			      $acq_stat_lookup, $c->{"GS_ID${i}"}, 
-			      $db_stats->{acq}, $db_stats->{acq_noid},
-			      $db_stats->{gui}, $db_stats->{gui_bad}, $db_stats->{gui_fail}, $db_stats->{gui_obc_bad},
-			      $db_stats->{avg_mag},
-			      $c->{"GS_ID${i}"});
-	  }
-	    else{
-                if ((not defined $db_handle) and ($c->{"TYPE${i}"} ne 'FID')){
```
taldcroft: The `not defined $db_handle` check looks to be gone now, and I assume that is on purpose.  I actually don't know exactly what that was doing, but were you going to update the `$acq_stat_lookup` to use the kadi server?  I'm not worried about waiting a few more seconds to get the kadi / mica page.  We want to go this direction to sever sybase dependencies.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1894,44 +1898,40 @@ sub print_report {
...
+                                         $db_stats->{acq}, $db_stats->{acq_noid},
+                                         $db_stats->{gui}, $db_stats->{gui_bad},
+                                         $db_stats->{gui_fail}, $db_stats->{gui_obc_bad},
+                                         $db_stats->{avg_mag})
+            }
+
+            if (($c->{"TYPE$i"} ne 'FID')){
+                my $cat_blurb = "";
+                if (defined $c->{"GS_MAGERR$i"}){
+                    $cat_blurb = sprintf("mac_aca_err=%4.2f(mags) pos_err=%4.2f(arcsec) color1=%4.2f <BR />",
```
taldcroft: The `(mags)` and `(arcsec)` are not necessary because these correspond to units in the rest of the table (Mag and Y/Z angle in arcsec).

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: Actually, I think the $db_handle thing here was a mistake.  If I cut that, I think we lose the "free" benefit of only having URLs to click for stars that have history.  I'll review and fix.

I opened a separate issue to update the acq_stat_lookup to use the kadi / mica page and was not planning to hit that here (I'd prefer to make that change when we promote the guide and acq stat lookups through the mica database so everything is fully consistent).

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: Gotcha.  Yes, I was afraid our future selves would thing these were in the AGASC table units unconverted.

taldcroft: Since the raw AGASC values are integers there shouldn't be a problem with ambiguity.

jeanconn:
Ugh.  I made the code worse again but it looks like I'm getting back the behaviors I want.  I'll try to figure out a cleaner way forward on this.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
taldcroft: If you are sticking with the CGI that does not include AGASC info (which is fine), then leaving this is OK.  For me what I want is the hover-over that tells me if the star has history, plus the always-available link to get the full AGASC info.  Agreed this is some merit to being able to see instantly at a top-level if a star has history, but I think always being able to easily get the AGASC info is more important.

jeanconn: Good point.  I'll add that to requirements in the issue for converting to the kadi star lookups.

PR #250, opened 2018-08-13
Updates for Perl 5.26.2, Py3, and modern hopper

jeanconn:
Updates for Perl 5.26.2 and Py3 for skare3

### DIFF
--- a/starcheck/pcad_att_check.py
+++ b/starcheck/pcad_att_check.py
@@ -121,22 +121,18 @@ def make_pcad_attitude_check_report(backstop_file, or_list_file=None, mm_file=No
     # gives the history of updates as a dict with a `value` and `date` key.
     sc = hopper.run_cmds(backstop_file, or_list, ofls_characteristics_file,
                          initial_state=initial_state)
-    # Iterate through obsids in order
-    obsids = [obj['value'] for obj in sc.obsids]
-    for obsid in obsids:
-        if obsid not in sc.checks:
-            continue
-
-        checks = sc.checks[obsid]
-        for check in checks:
-            if check['name'] == 'CheckObsreqTargetFromPcad':
-                ok = check['ok']
+    # Iterate through checks by obsid to print status
+    checks = sc.get_checks_by_obsid()
+    for obsid in sc.obsids:
+        for check in checks[obsid]:
+            if check.name == 'attitude_consistent_with_obsreq':
+                ok = check.success
                 all_ok &= ok
-                if check.get('skip'):
-                    message = 'SKIPPED: {}'.format(check['message'])
+                if check.not_applicable:
+                    message = 'SKIPPED: {}'.format(":".join(check.infos))
                 else:
-                    message = 'OK' if ok else check['message']
-                line = '{:5d}: {}'.format(obsid, message)
+                    message = 'OK' if ok else "ERROR: {}".format(":".join(check.errors))
+                    line = '{:5d}: {}'.format(obsid, message)
                 lines.append(line)
 
     if out is not None:
--- a/starcheck/pcad_att_check.py
+++ b/starcheck/pcad_att_check.py
@@ -1,9 +1,11 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import re
-import hopper
+from astropy.table import Table
+import Quaternion
+
 from parse_cm import read_backstop, read_maneuver_summary, read_or_list
 from Chandra.Time import DateTime
-from astropy.table import Table
+import hopper
 
 
 def check_characteristics_date(ofls_characteristics_file, ref_date=None):
--- a/starcheck/pcad_att_check.py
+++ b/starcheck/pcad_att_check.py
@@ -69,13 +69,17 @@ def make_pcad_attitude_check_report(backstop_file, or_list_file=None, mm_file=No
     lines = []  # output report lines
 
     mm = read_maneuver_summary(mm_file)
-    q = [mm[0][key] for key in ['q1_0', 'q2_0', 'q3_0', 'q4_0']]
+    q = Quaternion.normalize([mm[0][key] for key in ['q1_0', 'q2_0', 'q3_0', 'q4_0']])
     bs = read_backstop(backstop_file)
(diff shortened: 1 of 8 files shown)


starcheck/src/starcheck.pl
```diff
@@ -49,21 +49,43 @@
...
 
 import os
+import six
 from chandra_aca.star_probs import set_acq_model_ms_filter
 import starcheck
 from starcheck.pcad_att_check import make_pcad_attitude_check_report, check_characteristics_date
 from starcheck.calc_ccd_temps import get_ccd_temps
 from starcheck.version import version
 
+def debyte_kwargs(kwargs):
```
jeanconn: I could probably use a hand simplifying this @taldcroft .  Looks like when I pass a dictionary from Perl as an argument, all the string keys and values are now bytes, so here I've pretty much just looped over everything to decode.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -2102,7 +2102,7 @@ sub print_report {
 	else{
 	    $o .= sprintf("<TD><img align=\"top\" src=\"%s/down.gif\" ></TD>", 
 			  $self->{STARCHECK} );
-	    $o .= sprintf("<TD>NEXT </TD>", $self->{next}->{obsid});
+	    $o .= sprintf("<TD>NEXT </TD>");
```
jeanconn: I suppose I should just cut the NEXT completely if there isn't actually a next to go to...

starcheck/src/starcheck.pl
```diff
@@ -49,21 +49,43 @@
...
+    if six.PY2:
```
jeanconn: Did we decide we're ready to have the code by Py3 only?

taldcroft: Let's plan on Py3 only.

starcheck/src/starcheck.pl
```diff
@@ -49,21 +49,43 @@
...
+       return kwargs
+    else:
+       td = dict()
+       for key, val in six.iteritems(kwargs):
+           try:
```
taldcroft: This touches on a philosophical question of duck-typing: for any object that has a `decode()` method, is it the right thing to call it?  For this case I'd rather take an opt-in approach and handle the case we know about, which is `bytes`:
```
if isinstance(key, bytes):
    key = key.decode('ascii')
```
And make the encoding explicit for this use case, since otherwise it uses the system default which is not necessarily predictable.

taldcroft: BTW, an alternate one-line hack that I think works for arbitrary pure Python structures is:
```
In [11]: r = {b'name': {b'joe': [1, 'jim', b'joe']}}

In [12]: repr(r)
Out[12]: "{b'name': {b'joe': [1, 'jim', b'joe']}}"

In [13]: ast.literal_eval(re.sub(r"b'", "'", repr(r)))
Out[13]: {'name': {'joe': [1, 'jim', 'joe']}}
```
This does assume that the string `b'` never occurs naturally within a string, but I think that is the case in this application.

starcheck/pcad_att_check.py
```diff
@@ -121,22 +121,18 @@ def make_pcad_attitude_check_report(backstop_file, or_list_file=None, mm_file=No
...
+                ok = check.success
                 all_ok &= ok
-                if check.get('skip'):
-                    message = 'SKIPPED: {}'.format(check['message'])
+                if check.not_applicable:
+                    message = 'SKIPPED: {}'.format(":".join(check.infos))
                 else:
-                    message = 'OK' if ok else check['message']
-                line = '{:5d}: {}'.format(obsid, message)
+                    message = 'OK' if ok else "ERROR: {}".format(":".join(check.errors))
```
jeanconn: I'm not sure about the convention for using the hopper messages (errors, warnings, infos etc).

jeanconn:
At this point it looks like things are running, with the caveat that I'm getting that warning from the masked values in hopper 

```
Checking star catalog for obsid 19718
Checking star catalog for obsid 18676
/fido.real/skare3/lib/python3.6/site-packages/astropy/table/column.py:221: UserWarning: Warning: converting a masked element to nan.
  self_data = np.array(data, dtype=dtype, copy=copy)
Wrote HTML report to mar2017e.html
Wrote text report to mar2017e.txt
```

starcheck/src/lib/Ska/Starcheck/FigureOfMerit.pm
```diff
@@ -28,7 +28,18 @@ use Inline Python => q{
 
 from chandra_aca.star_probs import acq_success_prob, prob_n_acq, mag_for_p_acq
 
+def _mag_for_p_acq(p_acq, date, t_ccd):
+    try:
+       date = date.decode()
```
jeanconn: For these, if we're only Py3, I don't think I need to duck-type or even set a conditional on this.  date will always need a decode and it will always work.

starcheck/src/starcheck.pl
```diff
@@ -49,21 +49,28 @@
...
+    return get_ccd_temps(**debyte_dict(kwargs))
 
 def plot_cat_wrapper(kwargs):
+    kwargs = debyte_dict(kwargs)
     try:
         from starcheck.plot import make_plots_for_obsid
     except ImportError as err:
         # write errors to starcheck's global warnings and STDERR
         perl.warning("Error with Inline::Python imports {}\n".format(err))
+    kwargs['catalog'] = [debyte_dict(row) for row in kwargs['catalog']]
```
taldcroft: In theory this is not needed with the new `debyte_dict`.  In practice?

starcheck/src/starcheck.pl
jeanconn: I suppose I should also do more checking to confirm that I have no strings that end in `b'` as you note:
```
In [7]: r = {b'name': {b'joe': [1, 'jim', b'bob']}}

In [8]: ast.literal_eval(re.sub(r"b'", "'", repr(r)))
Out[8]: {'name': {'joe': [1, 'jim', 'bo']}}
```

jeanconn:
Ended up up redoing these in starcheck 12.0, so overcome by events.

PR #253, opened 2018-09-06
Use agasc 1p7 by default

jeanconn:
Switch the default agasc to 1p7.

The switch to use a new AGASC directory looks like it works with a few trivial edits on flight starcheck.

### DIFF
--- a/starcheck/src/starcheck.pl
+++ b/starcheck/src/starcheck.pl
@@ -96,7 +96,7 @@ my %par = (dir  => '.',
 		   );
 
 my $agasc_parent_dir = '/proj/sot/ska/data/agasc';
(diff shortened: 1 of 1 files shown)


jeanconn:
This is a keep-in-back-pocket commit, in case we don't get to moving to Python for the star fetching.  In fact the move to Python for star fetching might also come with a Py3 transition in starcheck and we're not quite there yet.

This is also not yet tested with the pure-Perl agasc interface.

taldcroft:
How did you verify that it is picking up AGASC 1.7?

jeanconn:
All I've verified is that it prints out it is using AGASC 1.7.  For a real functional test will probably need to pick out some stars that will have different 1.7 magnitudes and confirm they are right.

jeanconn:
overcome by events

PR #254, opened 2018-09-16
Changes to plug in Python agasc module to replace Ska::AGASC

jeanconn:
Changes to plug in Python agasc module to replace Ska::AGASC

### DIFF
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -21,6 +21,53 @@ package Ska::Starcheck::Obsid;
 use strict;
 use warnings;
 
+use Inline Python => q{
+
+from astropy.table import Table, Column
+import Quaternion
+from Ska.quatutil import radec2yagzag
+import agasc
+
+
+def _get_agasc_stars(ra, dec, roll, radius, date, agasc_file):
+    """
+    Fetch the cone of agasc stars.  Update the table with the yag and zag of each star.
+    Return as a dictionary with the agasc ids as keys and all of the values as
+    simple Python types (int, float)
+    """
+    stars = agasc.get_agasc_cone(float(ra), float(dec), float(radius), date, agasc_file)
+    q_aca = Quaternion.Quat([float(ra), float(dec), float(roll)])
+    yags, zags = radec2yagzag(stars['RA_PMCORR'], stars['DEC_PMCORR'], q_aca)
+    yags *= 3600
+    zags *= 3600
+    stars['yang'] = yags
+    stars['zang'] = zags
+
+    # Get a dictionary of the stars with the columns that are used
+    # This needs to be de-numpy-ified to pass back into Perl
+    stars_dict = {}
+    for star in stars:
+        stars_dict["{}".format(star['AGASC_ID'])] = {
+            'id': int(star['AGASC_ID']),
+            'class': int(star['CLASS']),
+            'ra': float(star['RA_PMCORR']),
+            'dec': float(star['DEC_PMCORR']),
+            'mag_aca': float(star['MAG_ACA']),
+            'bv': float(star['COLOR1']),
+            'color1': float(star['COLOR1']),
+            'mag_aca_err': float(star['MAG_ACA_ERR']),
+            'poserr': float(star['POS_ERR']),
+            'yag': float(star['yang']),
+            'zag': float(star['zang']),
+            'aspq': int(star['ASPQ1']),
+            'var': int(star['VAR']),
+            'aspq1': int(star['ASPQ1'])}
+
+    return stars_dict
+
+};
+
+
 use List::Util qw( max );
 use Quat;
 use Ska::ACACoordConvert;
@@ -33,7 +80,6 @@ use Ska::Convert qw(date2time);
 use Ska::Starcheck::FigureOfMerit qw( make_figure_of_merit set_dynamic_mag_limits );
 use RDB;
 
-use Ska::AGASC;
 use SQL::Abstract;
 use Ska::DatabaseUtil qw( sql_fetchall_array_of_hashref );
 use Carp;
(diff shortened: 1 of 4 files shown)


starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -21,6 +21,53 @@ package Ska::Starcheck::Obsid;
 use strict;
 use warnings;
 
+use Inline Python => q{
+
+from astropy.table import Table, Column
```
jeanconn: This can go.  From an earlier draft.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -21,6 +21,53 @@ package Ska::Starcheck::Obsid;
...
+    zags *= 3600
+    stars['yang'] = yags
+    stars['zang'] = zags
+
+    # Get a dictionary of the stars with the columns that are used
+    # This needs to be de-numpy-ified to pass back into Perl
+    stars_dict = {}
+    for star in stars:
+        stars_dict["{}".format(star['AGASC_ID'])] = {
+            'id': int(star['AGASC_ID']),
```
jeanconn: This could get cleaned up with something that just matches the columns to functions to convert to plain Python types, but there weren't that many columns (to just write it out this way).

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -2181,52 +2227,26 @@ sub get_agasc_stars {
 #############################################################################################
 
     my $self = shift;
-    my $AGASC_DIR = shift;
+    my $AGASC_FILE = shift;
```
jeanconn: lower case

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -2181,52 +2227,26 @@ sub get_agasc_stars {
...
-					     $q_aca);
-	$yag *= $r2a;
-	$zag *= $r2a;
-	if ($star->mag_aca() < -10 or $star->mag_aca_err() < -10) {
-	    push @{$self->{warn}}, sprintf("$alarm Star with bad mag %.1f or magerr %.1f at (yag,zag)=%.1f,%.1f\n",
-					   $star->mag_aca(), $star->mag_aca_err(), $yag, $zag);
+    foreach my $star (values %{$self->{agasc_hash}}) {
+	if ($star->{'mag_aca'} < -10 or $star->{'mag_aca_err'} < -10) {
+	    push @{$self->{warn}}, sprintf(
+                "$alarm Star with bad mag %.1f or magerr %.1f at (yag,zag)=%.1f,%.1f\n",
```
jeanconn: Should try to confirm warning still works

starcheck/src/starcheck.pl
```diff
@@ -1308,13 +1277,9 @@ =head1 OPTIONS
...
 
-=item B<-agasc <agasc>>
-
-Specify version of agasc ( 1p4, 1p5, or 1p6 ).  Default is 1p6 .
-
-=item B<-agasc_dir <agasc directory>>
+=item B<-agasc_file <agasc>>
 
-Specify directory path to agasc.  Overrides -agasc option.
+Specify locate of agasc h5 file.  Default is /proj/sot/ska/data/agasc/agasc1p7.h5 .
```
jeanconn: Typo.  location of

starcheck/src/run_regress
```diff
@@ -5,13 +5,15 @@
 RunRegression()
 # Run regression tests.
 # Arg 1 is the AGASC version (e.g. 1p5)
-# Arg 2 is the fid characteristics name (e.g. fid_CHARACTERIS_FEB07)
+# Arg 2 is the AGASC file for h5
```
jeanconn: Technically, we need changes to run the regression tests if we want to support the new API in the test version of starcheck.  We don't *really* need this agasc_file change to have a home in the tests if we don't mind using agasc1p7.h5 when called from starcheck on these test dirs (which should be fine).

So we could just cut the agasc 1p5 tests and remove both the previous agasc_dir option and the new agasc_file option so everything runs at default values.

jeanconn:
@taldcroft  Do we want to add a check that the guide summary mag is a reasonable match for the magnitude from the AGASC?  As far as I can tell, we just have a check that the guide summary mag is at least 1.4 mags brighter than maxmag.  If we want a more specific check, that output would become a diff when we run against old data unless we either use a date-based assignment or just call out the agasc h5 file in the regression script (which I have in this first draft of the pull request).

If we don't care about adding a check that the guide summary mag matches the star mag from the agasc, we could just set up the regression tests to remove all the old 1p5 weeks and then simplify the calls to flight and test starcheck to just remove the explicit agasc setting.

jeanconn:
I added the "compare guide summary mag to agasc mag" test and set it to a small tolerance in fb8115d run my 4 way tests (SEP1718A with agasc 1p6 and 1p7 and SEP1718T with agasc 1p6 and 1p7).  Then I relaxed the test to just be a sanity check with a large operational tolerance (0.1 mags).

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: Looks like this last came up operationally in the OCT1402D products which are too old to run in modern starcheck (and I don't recall if the test here is looking for defects in the agasc that have since been repaired or if there is just the general concern that fields that have stars that have missing data are unreliable).  This won't get tested now.

taldcroft: Not testing now is fine.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
taldcroft: FYI you can use `.item()` on a numpy object to convert to the appropriate native Python type.  I believe `tolist()` would also work here.  It won't shorten this code much and maybe having the explicit types is fine anyway.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -21,6 +21,52 @@ package Ska::Starcheck::Obsid;
...
+    yags *= 3600
+    zags *= 3600
+    stars['yang'] = yags
+    stars['zang'] = zags
+
+    # Get a dictionary of the stars with the columns that are used
+    # This needs to be de-numpy-ified to pass back into Perl
+    stars_dict = {}
+    for star in stars:
+        stars_dict["{}".format(star['AGASC_ID'])] = {
```
taldcroft: More idiomatically written as `str(star['AGASC_ID'])`

PR #257, opened 2018-10-11
WIP: Re-prototype use of mica stats instead of sybase

jeanconn:


starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -65,6 +65,39 @@ def _get_agasc_stars(ra, dec, roll, radius, date, agasc_file):
...
+
+    acqs = ACQS[ACQS['agasc_id'] == int(agasc_id)]
+    time = float(time)
+    acq_5y = acqs[(acqs['guide_tstart'] >= (time - (5 * 365 * 86400)))
+                 & (acqs['guide_tstart'] <= time)]
+    ok_5y = (acq_5y['img_func'] == 'star') & (~acq_5y['ion_rad']) & (~acq_5y['sat_pix'])
+
+    guides = GUIDES[GUIDES['agasc_id'] == int(agasc_id)]
+
+    mags = acqs['mag_obs'][acqs['mag_obs'] != 0].tolist()
```
taldcroft: `np.concatenate([x, y, ..])` is a good function for your toolbox and works well here.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -65,6 +65,39 @@ def _get_agasc_stars(ra, dec, roll, radius, date, agasc_file):
...
+
+ACQS = mica.stats.acq_stats.get_stats()
+GUIDES = mica.stats.guide_stats.get_stats()
+
+
+def get_mica_stats(agasc_id, time):
+
+    acqs = ACQS[ACQS['agasc_id'] == int(agasc_id)]
+    time = float(time)
+    acq_5y = acqs[(acqs['guide_tstart'] >= (time - (5 * 365 * 86400)))
```
taldcroft: Why the 5 year time limit?  I think the two important pieces of information are:
- Best estimate of observed mag
- Whether actual p_acq is persistently very different from model p_acq, e.g. due to a nearby bright spoiler.

Did you have a different idea?

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: I started with np.concatenate.  I didn't seem to work if either x or y was empty.  I'll go back to its docs.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: I didn't "rethink" about this when re-implimenting this last night, and left this mismatch (5y for acq on noid from the original prototype, all data for guide) in this WIP.

For acq, what do you consider a useful way to measure whether p_acq is persistently very different from model p_acq?

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: I must have just has something else wrong when I blamed np.concatenate

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -65,6 +65,40 @@ def _get_agasc_stars(ra, dec, roll, radius, date, agasc_file):
...
+GUIDES = mica.stats.guide_stats.get_stats()
+
+
+def get_mica_stats(agasc_id, time):
+
+    time = float(time)
+
+    acqs = ACQS[(ACQS['agasc_id'] == int(agasc_id))
+               & (ACQS['guide_tstart'] < time)]
+    ok = (acqs['img_func'] == 'star') & (~acqs['ion_rad']) & (~acqs['sat_pix'])
```
jeanconn: It looks like in Py3 all these string columns are now bytestrings so this match on 'star' doesn't work.   Should try to fix this on the mica side.
```
In [23]: acqs['img_func']
Out[23]: 
array([b'star', b'NONE', b'star', b'star', b'star', b'NONE', b'star',
       b'star', b'star', b'star', b'star', b'star', b'star', b'star',
       b'star', b'star', b'star', b'star', b'star', b'star', b'star',
       b'star'], 
      dtype='|S7')
```

taldcroft: If you turn it into an astropy `Table` then the compare will work.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
taldcroft: The super-simple thing to do is just add all the `p_acq` and compare to the actual number that were acquired.  If the probability model is applicable to the star then those values should match.  For extreme cases it is easy, e.g. sum of p_acq is 19 (for e.g. 23 trials) and in fact only 3 acqs were successful.   

The trick is in establishing confidence intervals for not-so-obvious cases.  This is basically like summing bernoulli distributions, but I'm not sure offhand if there is a clean computational formalism.  One could always do Monte-Carlo simulation with a modest number of trials (a few hundred) to find cases of "very difference from the model".

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: Wondering if I should do that on the mica side.  Also wondering if we have other codes that are going to be unhappy with this out of pytables.

taldcroft: This is a key sticking point in Py3 migration, and one of the main drivers for https://github.com/astropy/astropy/pull/5700.  Takeaway is to use astropy `Table` when dealing with character data coming from HDF5.

taldcroft: Upstream fix to use `Table` in mica seems like the right answer.  Probably that will break less code than not doing it.  Maybe??  :smile:

jeanconn:
Though I'm not actually sure if we want to finish this up or do https://github.com/sot/ska-projects/issues/114

taldcroft:
I'm lost on what is happening with all the commits that seem unrelated to using mica stats instead of sybase.

jeanconn:
I don't think this PR needs detailed review, I think we just need a plan for how to use previous observation data to review a new observation (you started to form some ideas for acq in this thread) and then need to decide if that should happen in starcheck or via agasc supplement.

PR #258, opened 2018-10-23
Safe mode 2018283 prov

jeanconn:
This implements several changes in starcheck to support load review after the 2018283 safe mode:

  - Sets the "fractional" guide count metric to use temperature-scaled magnitude limits for the counting
  - Sets the acquisition probability threshold to more conservative values per Tom's suggested thresholds in memo "ACA planning requirements for loads going forward"
  - Uses the ACA model from 24-Oct-2018 in chandra_models 3.18.1

### DIFF
--- a/ska_bin_starcheck
+++ b/ska_bin_starcheck
@@ -1,4 +1,4 @@
- #!/bin/bash
+#!/bin/bash
 
 SKA_bin=`dirname $0`
 eval `${SKA_bin}/flt_envs -shell sh`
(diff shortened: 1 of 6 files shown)


taldcroft:
Awesome.  Approved for LR with FD waiver.

jeanconn:
I should mark this with a dev version and tag as well.

jeanconn:
Do we want to just approve the provisional acquisition probability limits for flight starcheck now and make them less conservative in a future release?

taldcroft:
Yes, that makes sense.

PR #259, opened 2018-11-09
 Update to ACA model approved SSAWG 7-Nov-2018

jeanconn:
Update to ACA model approved SSAWG 7-Nov-2018

This also finally updates the planning limit in characteristics to -10.2 (even if we'll be going to -9.5 after a couple of weeks).

The intent is to use this version of starcheck with xija >= 3.12 (and to test with chandra_aca >= 3.24).

### DIFF
--- a/starcheck/data/aca_spec.json
+++ b/starcheck/data/aca_spec.json
@@ -1,431 +1,454 @@
 {
     "bad_times": [
         [
-            "2011:187:12:00:00", 
+            "2011:187:12:00:00",
             "2011:191:00:00:00"
-        ], 
+        ],
         [
-            "2011:299:04:00:00", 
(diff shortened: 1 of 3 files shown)


jeanconn:
I note that the new model has many whitespace differences from the old model.  I didn't care.

jeanconn:
@taldcroft If you are fine with these changes, I can try to put out to mpweekly for today and finish the regression test output evaluation later today.

taldcroft:
Is there a driver to push on this today?

jeanconn:
If you don't want to push for today that's fine.  I thought we just always wanted the best thermal model in starcheck and that these updates are trivial.  I suppose if we review next week we are still lined up with the Matlab promotion (and not behind a week).

PR #260, opened 2018-11-19
Minimal updates from starcheck 11.29 to run on skare3

jeanconn:
Minimal updates from starcheck 11.29 to run on skare3

### DIFF
--- a/ska_bin_starcheck
+++ b/ska_bin_starcheck
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-SKA_bin=`dirname $0`
+SKA_bin=/proj/sot/ska3/flight/bin
 eval `${SKA_bin}/flt_envs -shell sh`
 exec ${SKA_ARCH_OS}/bin/starcheck "$@"
--- a/starcheck/pcad_att_check.py
+++ b/starcheck/pcad_att_check.py
@@ -121,22 +121,18 @@ def make_pcad_attitude_check_report(backstop_file, or_list_file=None, mm_file=No
     # gives the history of updates as a dict with a `value` and `date` key.
     sc = hopper.run_cmds(backstop_file, or_list, ofls_characteristics_file,
                          initial_state=initial_state)
-    # Iterate through obsids in order
-    obsids = [obj['value'] for obj in sc.obsids]
-    for obsid in obsids:
-        if obsid not in sc.checks:
-            continue
-
-        checks = sc.checks[obsid]
-        for check in checks:
-            if check['name'] == 'CheckObsreqTargetFromPcad':
-                ok = check['ok']
+    # Iterate through checks by obsid to print status
+    checks = sc.get_checks_by_obsid()
+    for obsid in sc.obsids:
+        for check in checks[obsid]:
+            if check.name == 'attitude_consistent_with_obsreq':
+                ok = check.success
                 all_ok &= ok
-                if check.get('skip'):
-                    message = 'SKIPPED: {}'.format(check['message'])
+                if check.not_applicable:
+                    message = 'SKIPPED: {}'.format(":".join(check.infos))
                 else:
-                    message = 'OK' if ok else check['message']
-                line = '{:5d}: {}'.format(obsid, message)
+                    message = 'OK' if ok else "ERROR: {}".format(":".join(check.errors))
+                    line = '{:5d}: {}'.format(obsid, message)
                 lines.append(line)
 
     if out is not None:
--- a/starcheck/pcad_att_check.py
+++ b/starcheck/pcad_att_check.py
@@ -1,9 +1,11 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import re
-import hopper
+from astropy.table import Table
+import Quaternion
+
 from parse_cm import read_backstop, read_maneuver_summary, read_or_list
 from Chandra.Time import DateTime
-from astropy.table import Table
+import hopper
 
 
 def check_characteristics_date(ofls_characteristics_file, ref_date=None):
--- a/starcheck/pcad_att_check.py
+++ b/starcheck/pcad_att_check.py
@@ -3,7 +3,7 @@ import re
 from astropy.table import Table
 import Quaternion
 
-from parse_cm import read_backstop, read_maneuver_summary, read_or_list
+from parse_cm import read_backstop, read_or_list
 from Chandra.Time import DateTime
 import hopper
 
@@ -60,7 +60,22 @@ def recent_sim_history(time, file):
                 return greta_time, int(value)
 
 
-def make_pcad_attitude_check_report(backstop_file, or_list_file=None, mm_file=None,
+def recent_attitude_history(time, file):
+    """
+    Read from the end of the a ATTITUDE history file and return the
+    first (last) time and value before the given time.  Specific
+    to ATTITUDE and transition history based on the regex for
+    parsing.
+    """
+    for line in reversed(open(file).readlines()):
+        match = re.match('^(\d+\.\d+)\s+\|\s+(\S+)\s+(\S+)\s+(\S+)\s+(\S+)\s*', line)
+        if match:
+            greta_time, q1, q2, q3, q4 = match.groups()
+            if (DateTime(greta_time, format='greta').secs < time):
+                return greta_time, float(q1), float(q2), float(q3), float(q4)
+
+
+def make_pcad_attitude_check_report(backstop_file, or_list_file=None, attitude_file=None,
                                     simtrans_file=None, simfocus_file=None,
                                     ofls_characteristics_file=None, out=None,
                                     dynamic_offsets_file=None,
@@ -72,9 +87,12 @@ def make_pcad_attitude_check_report(backstop_file, or_list_file=None, mm_file=No
     all_ok = True
     lines = []  # output report lines
 
-    mm = read_maneuver_summary(mm_file)
-    q = Quaternion.normalize([mm[0][key] for key in ['q1_0', 'q2_0', 'q3_0', 'q4_0']])
     bs = read_backstop(backstop_file)
+
+    # Get initial state attitude and sim position from history
+    att_time, q1, q2, q3, q4 = recent_attitude_history(DateTime(bs[0]['date']).secs,
+                                            attitude_file)
+    q = Quaternion.normalize([q1, q2, q3, q4])
     simfa_time, simfa = recent_sim_history(DateTime(bs[0]['date']).secs,
                                            simfocus_file)
     simpos_time, simpos = recent_sim_history(DateTime(bs[0]['date']).secs,
--- a/starcheck/pcad_att_check.py
+++ b/starcheck/pcad_att_check.py
@@ -142,7 +142,7 @@ def make_pcad_attitude_check_report(backstop_file, or_list_file=None, attitude_f
     # Any state value (e.g. obsid or q_att) has a corresponding plural that
     # gives the history of updates as a dict with a `value` and `date` key.
     sc = hopper.run_cmds(backstop_file, or_list, ofls_characteristics_file,
-                         initial_state=initial_state)
+                         initial_state=initial_state, starcheck=True)
     # Iterate through checks by obsid to print status
     checks = sc.get_checks_by_obsid()
     for obsid in sc.obsids:
--- a/starcheck/pcad_att_check.py
+++ b/starcheck/pcad_att_check.py
@@ -69,13 +69,17 @@ def make_pcad_attitude_check_report(backstop_file, or_list_file=None, mm_file=No
     lines = []  # output report lines
 
     mm = read_maneuver_summary(mm_file)
-    q = [mm[0][key] for key in ['q1_0', 'q2_0', 'q3_0', 'q4_0']]
+    q = Quaternion.normalize([mm[0][key] for key in ['q1_0', 'q2_0', 'q3_0', 'q4_0']])
     bs = read_backstop(backstop_file)
     simfa_time, simfa = recent_sim_history(DateTime(bs[0]['date']).secs,
                                            simfocus_file)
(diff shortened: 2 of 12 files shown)


starcheck/pcad_att_check.py
```diff
@@ -115,22 +121,18 @@ def make_pcad_attitude_check_report(backstop_file, or_list_file=None, mm_file=No
...
+    # Iterate through checks by obsid to print status
+    checks = sc.get_checks_by_obsid()
+    for obsid in sc.obsids:
+        for check in checks[obsid]:
+            if check.name == 'attitude_consistent_with_obsreq':
+                ok = check.success
                 all_ok &= ok
-                if check.get('skip'):
-                    message = 'SKIPPED: {}'.format(check['message'])
+                if check.not_applicable:
```
jeanconn: Not entirely sure if this was "right" using the newer hopper API, but I can't recall what the drivers were to the changes on the hopper side.

starcheck/src/starcheck
```diff
@@ -1,7 +1,12 @@
 #!/bin/bash
 
-unset LD_LIBRARY_PATH
-unset SYBASE
+# Don't unset LD_LIBRARY_PATH and SYBASE because 'perl' below
```
jeanconn: I don't think we need a 'perl3ska' or the like.  If this is always being called from another wrapper we control, we can call flt_envs and set SYBASE and LD_LIBRARY_PATH appropriately just before this.  If there is anything from the MP environment, it should just end up further down the path.

starcheck/src/lib/Ska/Starcheck/Dark_Cal_Checker.pm
```diff
@@ -1895,8 +1896,15 @@ package TLREntry;
 
 use strict;
 use Carp;
-use Ska::Convert qw(date2time);
```
jeanconn: I think that Ska::Convert date2time and time2date both have issues and I'd like to replace them globally with the Chandra.Time methods, but it looks like subbing in just at this location fixes some issues with unexpected delta times when checking that the timing of the commands matches the template in the dark call checker.

starcheck/pcad_att_check.py
```diff
@@ -67,13 +73,17 @@ def make_pcad_attitude_check_report(backstop_file, or_list_file=None, mm_file=No
...
-    q = [mm[0][key] for key in ['q1_0', 'q2_0', 'q3_0', 'q4_0']]
+    q = Quaternion.normalize([mm[0][key] for key in ['q1_0', 'q2_0', 'q3_0', 'q4_0']])
     bs = read_backstop(backstop_file)
     simfa_time, simfa = recent_sim_history(DateTime(bs[0]['date']).secs,
                                            simfocus_file)
     simpos_time, simpos = recent_sim_history(DateTime(bs[0]['date']).secs,
                                              simtrans_file)
-    initial_state = {'q_att': q,
+
+    initial_state = {'q1': q[0],
```
jeanconn: This is starting to get to a point that keeps hopper happy, but this is using the maneuver summary to set initial quats... which is not consistent with replan/reopen when the backstop passed to hopper starts before the DOT and maneuver summary.  And then we've got all of the back and forth about checking the pre-DOT interval.

Mostly I think we've been spared more errors from hopper here because the rest of the stuff that hopper is trying to do is not implemented (but it looks to be configured to fetch stars and try to build catalogs to check even though I don't think we are using any of those features).  Might be easiest just to comment out everything after the attitude checking and come back to hopper improvements.

jeanconn: Updated to set initial quats from ATTITUDE history file.

jeanconn:
With regard to the current differences in timing in the regression tests.  
First, Ska::Convert date2time is a disaster:  https://github.com/sot/Ska-Convert/blob/master/Convert.pm#L101

jeanconn:
Then, it looks to me like modern Perl in ska3 seems to do better than the Perl in ska

Python reference
```
In [1]: import Chandra.Time
In [2]: Chandra.Time.DateTime("2015:351:09:49:16.927").secs
Out[2]: 566733025.111
```

ska3 perl
```
ska3-fido: perl -e 'use Ska::Convert qw(date2time); print date2time("2015:351:09:49:16.927") . "\n"'
566733020.111
```

ska perl
```
ska-fido: perl -e 'use Ska::Convert qw(date2time); print date2time("2015:351:09:49:16.927") . "\n"'
566733019.184
```
I'm wondering if timegm is improved in ska3 perl?  It is looking to me in the first spot check that the the newer code has the more correct answer and that's giving us a regression diff.  I'll review the other two diffs.

jeanconn:
Doh.  I missed that we've got the .111 part right in ska3 perl but are still off by 5 seconds.

starcheck/pcad_att_check.py
taldcroft: Sigh, I don't recall specifically but this was part of a big refactor early on.

ska_bin_starcheck
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-SKA_bin=`dirname $0`
+SKA_bin=/proj/sot/ska3/flight/bin
```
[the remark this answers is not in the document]
jeanconn: My thought was that we would have keep the launcher this version of  starcheck in /proj/sot/ska/bin, so the `dirname $0` would be incorrect for determining SKA_bin .

starcheck/src/starcheck.pl
```diff
@@ -54,16 +54,25 @@
 from starcheck.calc_ccd_temps import get_ccd_temps
 from starcheck.version import version
 
+# Borrowed from https://stackoverflow.com/a/33160507
```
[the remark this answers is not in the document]
jeanconn: I'm a little confused about what subclasses you expect will be passed back from Perl?  I thought even handling `set` seemed a stretch.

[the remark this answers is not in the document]
taldcroft: It's just a habit from astropy to always consider subclasses in any code. Agreed it doesn't have any impact in this narrow use case but I thought this function might be useful elsewhere.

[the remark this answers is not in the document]
jeanconn: Gotcha.  Yes, we may end up wanting something like this in a utils module for Perl.  I think this is passable now though, so I'm going to try to run the regression tests again for record (not actually sure if we'll make a load review with running them at least twice more).

jeanconn:
I think I'd like to leave this tagged as 12.0 and consider the probability threshold update (and integration with the updated chandra_aca and agasc) as 12.1 .

jeanconn:
Looks like I'm getting a spurious warning on part of the Large Dither checks now because the times are too accurate (I think we passed the timing test before a bit by accident).  I'll have at least one more change to Obsid.pm for this.  Also while looking at that, not as sure now about why we are subtracting 10 seconds from the maneuver summary times (some "offset" is assumed in there but I've forgotten if that is relative to the DOT or Backstop, or if it is with regard to AONMMODE or AOMANUVR).

taldcroft:
I have no concerns and you can proceed with the plan you outlined.

I don't recall the specifics of the 10 second offset, but I do remember discussing this at a SSAWG.  My recollection is that there is really an offset in the man summary file relative to the commands (backstop), but that recollection is only at moderate confidence.  I'm more confident that the 10 seconds was there on purpose and is correct.

jeanconn:
I'm confident it was in there on purpose but "correct" is only correct if I know what it is supposed to mean when implementing or reviewing tests (and I looks like maybe I did not for the large dither tests; no big deal there though).  I'll plan to isolate my changes to the large dither test for now to keep change scope small.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -809,16 +810,13 @@ sub large_dither_checks {
...
-    my $time_tol = 10;         # Commands must be within $time_tol of expectation
+    my $time_tol = 11;         # Commands must be within $time_tol of expectation
 
     # Save the number of warnings when starting this method
     my $n_warn = scalar(@{$self->{warn}});
 
-    # The obs tstart and tstop times are derived from the OFLS maneuver summary file,
-    # which for implementation reasons gives the maneuver times offset by 10 seconds
-    # from the actual command timing. Fix this here.
-    my $obs_tstart = $self->{obs_tstart} - 10;
```
jeanconn: It looks like the maneuver summary parsing is adding 10 seconds to tstart but not tstop?

https://github.com/sot/starcheck/blob/08cf90fbb57534c126bc5111fa0b36ab7c5deb26/starcheck/src/lib/Ska/Parse_CM_File.pm#L724

Either way, subtracting it off this way was not working on this test against big dither commanding in load explicitly verified by PCAD.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -860,8 +858,8 @@ sub large_dither_checks {
         }
     }
     # Check that the dither state at the end of the observation started 5 minutes before
-    # the end (within time_tol)
-    if ((abs($obs_tstop - $obs_stop_dither->{time} - 300) > $time_tol)){
+    # the end (within time_tol) .  obs_tstop appears not corrected by 10s so use 310 instead of 300
+    if ((abs($obs_tstop - $obs_stop_dither->{time} - 310) > $time_tol)){
```
jeanconn: One extra pad seems to work at the end of the observation (so, using 310 instead of 300 for this test).

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -809,16 +810,13 @@ sub large_dither_checks {
     my $self = shift;
     my $dither_state = shift;
     my $all_dither = shift;
-    my $time_tol = 10;         # Commands must be within $time_tol of expectation
+    my $time_tol = 11;         # Commands must be within $time_tol of expectation
```
jeanconn: Increased time tolerance by 1s.  My impression for these big dither tests is that a change in the tolerance from 10 to 11 (or possible 21 when including extra 10 second pad ) is fine.

taldcroft: Agreed.

jeanconn: The large dither checks are somewhat broken anyway, because we have the defect in flight starcheck that the operational process has changed (disable dither to start not required) and the code has not caught up.

jeanconn:
Thanks!  And yes, I'd come to that conclusion with regard to AONMMODE and AOMANUVR vs the maneuver summary, but for the end of the big dither for example, I wasn't sure which was more appropriate as the reference (need to enable normal dither 5 minutes before the next maneuver or 5 minutes before the next maneuver mode transition?).

With regard to maneuver stop time, I suspect we should subtract 10 seconds from the parsed maneuver summary stop time as well in Parse_CM_FIles for consistency, but as mentioned, didn't want this change to get out of scope.

jeanconn:
There's one diff now that isn't completely obvious

```
********************* 2011/APR0411/oflsa ********************
--- /home/jeanconn/git/starcheck3/test_regress/release/2011/APR0411/oflsa/starcheck.txt                                           2018-11-23 14:33:49.308805000 -0500
+++ /home/jeanconn/git/starcheck3/test_regress/han-v.cfa.harvard.edu_f0a67573c9dd2fdf22f318d3eb9e3ebcb06cc0fa/2011/APR0411/oflsa/st
archeck.txt	2018-11-28 22:03:11.043897000 -0500
@@ -8,6 +8,7 @@
 Using TLR file /data/mpcrit1/mplogs/2011/APR0411/oflsa/CR093_2200.tlr
 Using acq_star_rdb file ${DATA}/bad_acq_stars.rdb
 Using agasc_file file /proj/sot/ska/data/agasc/agasc1p6.h5
+Using attitude file /data/mpcrit1/mplogs/2011/APR0411/oflsa/History/ATTITUDE.txt
 Using backstop file /data/mpcrit1/mplogs/2011/APR0411/oflsa/CR093_2200.backstop
 Using bad_pixel file ${DATA}/ACABadPixels
 Using banned_agasc file ${DATA}/agasc.bad
@@ -256,9 +257,9 @@
 
 >> WARNING: [11] Backstop YAG,ZAG differs from AGASC by > 1.5 arcsec: dyag = 1.73 dzag = 1.28 
 
-Probability of acquiring 2,3, and 4 or fewer stars (10^x):	-9.176	-7.107                                                    -5.244	
+Probability of acquiring 2,3, and 4 or fewer stars (10^x):	-9.183	-7.114                                                    -5.250	
 Acquisition Stars Expected  : 7.86
-Predicted Max CCD temperature: -19.3 C 	 N100 Warm Pix Frac 0.071
+Predicted Max CCD temperature: -19.2 C 	 N100 Warm Pix Frac 0.071
 Dynamic Mag Limits: Yellow 10.30 	 Red 10.60
```


[regress_diffs.txt](https://github.com/sot/starcheck/files/2628845/regress_diffs.txt)
[vehicle_regress_diffs.txt](https://github.com/sot/starcheck/files/2628846/vehicle_regress_diffs.txt)

taldcroft:
All looks good!  :tada:

For the record, from:
https://occweb.cfa.harvard.edu/twiki/bin/view/MissionPlanning/LargeOffNominalDither

"Once backstop has been run, be sure to check the backstop file to verify that the 5 minutes exists between the AOENDITH and AOMANUVR commands"

jeanconn:
Thanks for the dither reference!  I was looking in the guidelines by category and didn't find it.

taldcroft:
> I was looking in the guidelines by category and didn't find it.

I started there as well, but persisted.  :smile:

taldcroft:
And the single :tada: doesn't fully capture how awesome this is.

jeanconn:
Thanks Tom.  I note that I've been pushing hard on this on Linux, but haven't built or tested on OSX in a while.

PR #262, opened 2018-11-29
Updates to accompany grid-acq-prob promotion

jeanconn:
Updates to accompany the acquisition probability model in chandra_aca 4.24

   - Reverts provisional acquisition probability thresholds (sets back to P2=~2.1)
   - References the -9.5C planning limit

### DIFF
--- a/starcheck/data/characteristics.yaml
+++ b/starcheck/data/characteristics.yaml
@@ -60,8 +60,8 @@ no_starcat_oflsid:
    - RDE
    - DC_T
 
-ccd_temp_yellow_limit: -11.5
-ccd_temp_red_limit: -10.2
+ccd_temp_yellow_limit: -10.2
+ccd_temp_red_limit: -9.5
 
 n100_warm_frac_default: .15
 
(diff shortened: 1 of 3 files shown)


jeanconn:
(Re)-testing this with chandra_aca 4.24 and agasc 4.7 .  Starcheck regression outputs will be big for the probability change.

starcheck/data/characteristics.yaml
```diff
@@ -60,8 +60,8 @@ no_starcat_oflsid:
    - RDE
    - DC_T
 
-ccd_temp_yellow_limit: -11.5
-ccd_temp_red_limit: -10.2
+ccd_temp_yellow_limit: -10.2
```
jeanconn: I think the yellow limit is just used to plot another line on the ccd temperature plot at the top... which isn't used for anything.  Should probably cut that from the plotting code, but for now it needs to be set to something.

jeanconn:
So the regress diffs appear to be ~4Mb.

jeanconn:
The regress log verifying that it actually ran on all the loads is more relevant for the PR.  The diffs just show that the all the star probabilities have changed.  

[regress_log.txt](https://github.com/sot/starcheck/files/2635191/regress_log.txt)

PR #266, opened 2018-12-12
Two-axis dither, different guide/acq dither, and kadi continuity

jeanconn:
Updates to handle two-axis dither, different guide/acq dither, and kadi continuity


starcheck/src/lib/Ska/Parse_CM_File.pm
```diff
@@ -103,17 +93,33 @@ sub dither {
...
-    
-    my @ok = grep { $dh_time[$_] < $bs_arr->[0]->{time} } (0 .. $#dh_time);
-    my @state = (@dh_state[@ok], @bs_state);
-    my @time   = (@dh_time[@ok], @bs_time);
-    my @params = (@dh_params[@ok], @bs_params);
+
+    # Confirm that last state matches kadi continuity ENAB/DISA
+    if ($kadi_dither->{'dither'} ne $dh_state[-1]){
+        print STDERR "Dither status in kadi commands does not match DITHER history\n";
+        return (1, undef);
```
jeanconn: I there's something very weird, everything will end up getting checked with (20, 20) dither and a red warning about dither.

starcheck/src/lib/Ska/Parse_CM_File.pm
```diff
@@ -137,17 +153,32 @@ sub dither {
...
-                      period_y => $dither_period_y,
-                      tlmsid => $params[$_]->{TLMSID},
-                  };
+      if ($dither_state eq 'DISA'){
+          # If disabled, reset the amplitudes to be 0.  The params may be nonzero onboard
+          # but we're more interested in the effective amplitudes for starcheck.
+          push @dither, { time => $time[$_],
+                          state => 'DISA',
+                          source => $time[$_] < $bs_start ? 'kadi' : 'backstop',
+                          ampl_p => 0,
```
jeanconn: I think this is the only debatable change.  It just simplifies things for me later if I don't have to check dither status as well as amplitude.

starcheck/src/lib/Ska/Parse_CM_File.pm
jeanconn: This isn't really an additional check on kadi dynamic states, it is more of a check that the "cut" version from continuity is all as expected.  The defaults really aren't bad either if we end up with a rapid replan and an out-of-date kadi or some such... though I don't know if we want to be able to provide a state manually.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -201,7 +201,8 @@ sub set_ACA_bad_pixels {
 	#cut out the quadrant boundaries
 	foreach my $i ($line[0]..$line[1]) {
 	    foreach my $j ($line[2]..$line[3]) {
-		my $pixel = {};
+		my $pixel = {'row' => $i,
+                             'col' => $j};
```
jeanconn: This is not strictly related to dither, but helps with an improved warning message for bad pixels.

starcheck/src/lib/Ska/Parse_CM_File.pm
```diff
@@ -137,17 +153,32 @@ sub dither {
...
-                      period_p => $dither_period_p,
-                      period_y => $dither_period_y,
-                      tlmsid => $params[$_]->{TLMSID},
-                  };
+      if ($dither_state eq 'DISA'){
+          # If disabled, reset the amplitudes to be 0.  The params may be nonzero onboard
+          # but we're more interested in the effective amplitudes for starcheck.
+          push @dither, { time => $time[$_],
+                          state => 'DISA',
+                          source => $time[$_] < $bs_start ? 'kadi' : 'backstop',
```
taldcroft: I don't understand how `$time[$_] < $bs_start` can ever be true here.  This section of code is strictly processing backstop commands, so having this conditional here just makes me question if I understand what's going on.  In fact it did make me spend a lot more time re-reading the code, but that is not necessarily the goal.

starcheck/src/lib/Ska/Parse_CM_File.pm
```diff
@@ -103,17 +93,33 @@ sub dither {
...
+
+    my @state;
+    my @time;
+    my @params;
+    # Then get everything from backstop
+    foreach $bs (@{$bs_arr}) {
+	if ($bs->{cmd} =~ '(COMMAND_SW|MP_DITHER)') {
+	    my %params = %{$bs->{command}};
+	    if ($params{TLMSID} =~ 'AO(DSDITH|ENDITH|DITPAR)') {
+		push @state, $dith_cmd{$1};
```
taldcroft: `@state` is a particularly confusing variable name.  This is really the dither enable/disable status, whereas "state" normally means a collection of correlated state values at a time (which is `@dither` later).  I think you could call it `dither_enab`, and rename `dither_cmd` to `dither_enab_cmd_map`.

starcheck/src/lib/Ska/Parse_CM_File.pm
```diff
@@ -65,30 +65,20 @@ sub dither {
 ###############################################################
     my $dh_file = shift;      # Dither history file name
     my $bs_arr = shift;               # Backstop array reference
+    my $kadi_dither = shift;
```
taldcroft: Add a comment about where `kadi` comes from and the data structure.  Also add a comment for the whole function about what it does.  This function is really much more than parsing CM files.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -695,63 +686,52 @@ sub check_dither {
...
+
+    # Check for standard and large dither
+    # This needs to be before the checking for dither during the observation, because the pad
+    # for that checking is set here
+    if ($guide_dither->{state} eq 'ENAB'){
+        if ((not standard_dither($guide_dither)) or not standard_dither($acq_dither)){
             push @{$self->{yellow_warn}}, "$alarm Non-standard dither\n";
-            if ($dither->{ampl_y} > $large_dith_thresh or $dither->{ampl_p} > $large_dith_thresh){
-                $self->large_dither_checks($dither, $dthr);
+            if ($guide_dither->{ampl_y} > $large_dith_thresh or $guide_dither->{ampl_p} > $large_dith_thresh){
```
taldcroft: You had expressed concern about the silly "standard dither" concept, and I agree.  You should fully decouple "standard" dither from "large" dither in all tests.  So move this test for large dither out of the standard dither test block.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -695,63 +686,52 @@ sub check_dither {
...
+                $self->large_dither_checks($guide_dither, $dthr);
                 # If this is a large dither, set a larger pad at the end, as we expect
                 # standard dither parameters to be commanded at 5 minutes before end,
                 # which is greater than the 3 minutes used in the "no dither changes
                 # during observation check below
                 $obs_end_pad = 5.5 * 60;
             }
         }
+        else{
```
taldcroft: Likewise move this test out of the `else`.  Sure, given the current definition of `standard_dither()` this test cannot fail for both guide and acq dither being "standard", but it just makes the code logic more difficult to follow and more fragile (what if 20x20 becomes a standard dither for ACIS?  Could happen to support dynamic background).

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -841,11 +858,12 @@ sub large_dither_checks {
...
-    # Is dither disabled at EOM and one minute before?
-    if ((abs($obs_tstart - $obs_start_dither->{time} - 60) > $time_tol)
-            or ($obs_start_dither->{state} ne 'DISA')){
+
+    my $det = (($self->{SI} eq 'HRC-S') or ($self->{SI} eq 'HRC-I')) ? 'hrc' : 'acis';
+    # Is dither nominal for detector at EOM
+    if ($det ne standard_dither($obs_start_dither)){
         push @{$self->{warn}},
-            sprintf("$alarm Dither should be disabled 1 min before obs start for Large Dither\n");
+            sprintf("$alarm Dither should be detector nominal 1 min before obs start for Large Dither\n");
```
taldcroft: Does it need to be `detector nominal` or just not large?  The large dither rules are about PCAD and ACA, so I don't think the detector matters.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -768,6 +748,42 @@ sub check_dither {
...
+        # ACA-002
+        push @{$self->{warn}}, "$alarm Dither mismatch - OR: $or_val != Backstop: $bs_val\n"
+            if ($or_val ne $bs_val);
+    }
+    else{
+        push @{$self->{warn}},
+            "$alarm Unable to determine dither from OR list\n";
+    }
+
+    # If dither is anabled according to the OR, check that parameters match OR vs Backstop
```
taldcroft: "enabled"

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -781,23 +797,24 @@ sub standard_dither{
     my %standard_z_dither = (20 => 768.6,
                              8 => 707.1);
 
+    my $z_amp = int($dthr->{ampl_p} + .5);
```
taldcroft: 0.5 please.

I also just googled for sec to see if perl has a built-in round() function.  Apparently not.  Wow.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: I think it needs to be *a* nominal pattern (not large but with the wrong period would be another issue).  AFAIK anything other than detector nominal dither would speak to either the dither states getting screwed up or DOT manual edits when they should not be required in ORViewer.  Might be safest to just confirm it is detector-nominal dither.

taldcroft: Another topic for SSAWG.  I don't see why 4x4 dither leading up to a large dither observation would be a problem.  The perennial question, are we checking that the loads are safe for the spacecraft or are we checking the planning process?

jeanconn: In this case, I still think it is easier to check the planning process.  But I can add a check on rates.

taldcroft: You don't need to check rates, that is PCAD not ACA.  Just amplitude, which is what matters for ACA.

jeanconn: For ACA we don't care at all if the proper dither is used for the proseco.acq .

But starcheck output is also used as a PCAD extra check in other areas as well.

starcheck/src/lib/Ska/Parse_CM_File.pm
jeanconn: Good catch.  I think this was a benign error on my part when I repurposed the previous code.  Can be simplified.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: It is trickier to "fully decouple" this when it is easiest to use the obs_end_pad hack.

starcheck/src/lib/Ska/Parse_CM_File.pm
```diff
@@ -61,93 +61,120 @@ sub TLR_load_segments{
...
+                    'source' => 'kadi',
+                    'ampl_p' => $dither_ampl_p,
+                    'ampl_y' => $dither_ampl_y,
+                    'period_p' => $dither_period_p,
+                    'period_y' => $dither_period_y};
+
+
+
+    # Build dither states using just kadi initial state and backstop cmds
+    foreach $bs (@{$bs_arr}) {
```
jeanconn: There was a loop over the bs_arr that was just making three new arrays and then looping over '@states' from that.  I didn't really see the point in looping twice (except to save one level of indentation in the larger block).

PR #271, opened 2019-01-05
Clip t_ccd and halfwidth 

jeanconn:
Set to clip t_ccd and halfwidth (and add yellow warnings) instead of passing those out-of-range values to star_probs.  This prevents the UserWarnings at console.


jeanconn:
OBE in #272

PR #272, opened 2019-01-05
Use proseco for P_ACQ

### LINKED ISSUE #268: Decide what to do with grid-acq model box warnings
```
Checking star catalog for obsid 48613
/proj/sot/ska3/flight/lib/python3.6/site-packages/chandra_aca/star_probs.py:297: UserWarning: 
Model grid-floor-2018-11 computed between halfw <= 60.0 <= 180.0, clipping input halfw(s) outside that range.
  .format(model, name, val_lo, val_hi, name))
```

### LINKED ISSUE #283: Set a more conservative P2 red limit for ERs
Use P2 3.0 as the red limit for ERs.

Use P2 3.0 as a yellow limit for ORs.

jeanconn:
Use proseco for P_ACQ

Closes #268 
Closes #283


starcheck/src/lib/Ska/Parse_CM_File.pm
```diff
@@ -747,6 +747,11 @@ sub MM {
         $manvr_hash{q2} = $quat[1];
         $manvr_hash{q3} = $quat[2];
         $manvr_hash{q4} = $quat[3];
+        my @init_quat = ($1,$2,$3,$4) if ($initial_attitude =~ /Quaternion:\s+(\S+)\s+(\S+)\s+(\S+)\s+(\S+)/);
```
jeanconn: I don't recall if I had a reason to directly calculate the maneuver angle; else we can just use the angle from the maneuver summary too.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -2639,7 +2638,7 @@ sub check_big_box_stars{
         }
     }
     if ($big_box_count < 3){
-        push @{$self->{warn}}, "$alarm Fewer than 3 ACQ stars with boxes >= 160 arcsec\n";
+        push @{$self->{yellow_warn}}, "$alarm Fewer than 3 ACQ stars with boxes >= 160 arcsec\n";
```
jeanconn: Should I just chuck this warning altogether for starcheck 13.0?  I thought we might want training wheels for one release.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -2656,14 +2655,195 @@ sub set_ccd_temps{
...
+        halfwidths => \@halfwidths,
+        fids => \@fid_ids,
+        guides => \@gui_ids,
+        acq_indexes => \@acq_indexes);
+
+    return \%proseco_args
+
+}
+
+my $CUM_PROB_LIMIT = 0.008;
```
jeanconn: Did you want just P2=2.1 or P2=2.0 for starcheck 13.0 or a new value?

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -686,6 +685,11 @@ sub check_dither {
     my $obs_tstart = $self->{obs_tstart};
     my $obs_tstop = $self->{obs_tstop};
 
+    unless (defined $obs_tstart){
```
jeanconn: This is an unrelated bugfix, but it was coming up the replan/reopn weeks I was using to look at probabilities, and it is harder to check that the probability fixes are right when there is a bunch of useless error reporting at console.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -72,7 +72,7 @@ def _get_agasc_stars(ra, dec, roll, radius, date, agasc_file):
 };
 
 
-use List::Util qw(max min);
+use List::Util qw(min max);
```
jeanconn: Hah.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
taldcroft: Chuck it.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
taldcroft: P2 = 2.0 (0.01).

starcheck/calc_ccd_temps.py
```diff
@@ -210,6 +210,7 @@ def get_interval_data(intervals, times, ccd_temp, obsreqs=None):
             obstemps[str(interval['obsid'])] = obs
             continue
         obs['ccd_temp'] = np.max(ok_temps)
+        obs['ccd_temp_acq'] = np.max(ok_temps[0:2]) if len(ok_temps) > 1 else obs['ccd_temp']
```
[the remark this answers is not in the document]
jeanconn: Do you mean I can cut the if statement?  or just change ok_temps[0:2] to ok_temps[:2]? 
I think the if state it was there for when ok_temps was empty in which case np.max gave nan?

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1998,9 +1997,13 @@ sub print_report {
 
             my $acq_prob = "";
             if ($c->{"TYPE$i"} =~ /BOT|ACQ/){
-                my $prob = $self->{acq_probs}->{$c->{"IMNUM${i}"}};
-                $acq_prob = sprintf("Prob Acq Success %5.3f", $prob);
-
+                if (defined $self->{acq_probs}->{$c->{"IMNUM${i}"}}){
```
taldcroft: Why this new test for existence?  Don't we always have an acq prob for any acq star?  Anyway couldn't you do something like `my $prob = $self->{acq_probs}->{$c->{"IMNUM${i}"}} or 0.0`?  (If that works Perl).

jeanconn:
Regarding  "I just started looking at this. Since this is new code I'd like to see a lot more code comments and function docstrings before I review."   Thanks that's fine.  I just didn't want to get too far down any path only to find out you hated a concept / strategy.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: I think your "or" version is good.  I keep trying to anticipate "unitialized" errors but I guess you don't get one on the hash lookup in the "or" version.

With regard to why this new test for existence, I didn't think a comment would help here (when the information about the *change* wouldn't be useful to me in the future) and stuck it in the commit message ("Handle missing data (ReOpen schedule beginning) for proseco probs").  We have a bunch of missing probabilities for stars at the beginning of a replan/reopen (old probs were calculated on those stars, new probs are not).   I could just not print the observations but...

taldcroft: I can't emphasize enough that code comments about why a particular test is done are *extremely valuable* to me and to your future self.  Basically during development you had some non-obvious insight into a particular situation that needs handling.  So explain that in detail *in the code* at the moment you think of it.  When we look at that line in 5 years we won't have to wonder.

You should not consider commit messages as a form of legacy documentation.  They provide some benefit during the code review process (though honestly I don't pay too much attention) but not so much later on.  Git blame does work and in some cases helps, but I would rather just see comments right in the code.

jeanconn: I just haven't put as much time into trying to figure out the Perl.  I'm playing whack-a-mole with uninitialized warnings, and sometimes I even go crazy and change two things at once. If the warning goes away I stop playing.  That's why I thought comments wouldn't be much use to my future self.  

In this case, I thought your idea for the "or" worked on the undefined quantity, but now I'm seeing   "Use of uninitialized value $prob in sprintf at starcheck/src/lib/Ska/Starcheck/Obsid.pm line 2009." again so I'm going to go back to checking to see if the specific image probs is defined.  Probably don't have to set to to 0 for the hover-over though, can leave it as the empty string.

taldcroft: I can't see any reason to waste time trying to make the code work for the replan/reopen section.  I have zero confidence in any of the starcheck results produced there and I never look at those, so you should just cut it.

Of course that would be hard in the context of these changes right now, so take that as something to do later and then clean up this code.

taldcroft: The thing about comments: the more you struggle with a section of code and the more moles you whacked, the more important it is that you document that struggle because later on someone (e.g. me) can come along with a brilliant idea to tidy up the code and have no idea of the back-story of how it got there and what will break if I make some little change that seems reasonable.

jeanconn: Confused by the comment, as I didn't even try to make the proseco probabilities get calculated for the replan/reopen observations (I started with that but stopped when you said so). I just made starcheck not spit out any *new* "uninitialized" warnings for those observations.  Since they were set up differently, that required some more handling and coding.  If we want to not print out *anything* in that section of the schedule, that's a different question.

starcheck/calc_ccd_temps.py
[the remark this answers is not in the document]
taldcroft: Yes, cut the if statement.  If `ok_temps` is empty then `np.max` gives an exception, not NaN.  So if the statement above has not crashed, then `np.max(ok_temps[:2])` will succeed.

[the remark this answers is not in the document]
jeanconn: OK.  I'll retest.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: Also, my confusion was for your comment one up (in case there was still some confusion about replan reopen and proseco probabilities).

 Regarding comments in general, I think in most of this code it will break if you make some little change that seems reasonable; but I'm not arguing about adding comments, just stating that there's more work to doing the investigation to make a truly useful comment and I have no idea how much time to spend on any of the Perl.  I thought I was being somewhat judicious in the use of my time, given present time challenges, but will try to find a more verbose compromise.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -2656,14 +2634,235 @@ sub set_ccd_temps{
...
+# This assigns the individual acq star probabilites back into $self->{acq_probs} and
+# assigns the P2 and expected values into $self->{figure_of_merit}.
+    my $self = shift;
+    my $cat_cmd = find_command($self, "MP_STARCAT");
+    my $args = $self->{proseco_args};
+
+    if (not %{$args}){
+        return;
+    }
+    my ($p_acqs, $two_or_fewer, $expected) = proseco_probs($args);
```
taldcroft: Call this `$P2` to stick with the new lingo and your comments.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -2148,14 +2151,10 @@ sub print_report {
...
     if (exists $self->{figure_of_merit}) {
-	my @probs = @{ $self->{figure_of_merit}->{cum_prob}};
 	my $bad_FOM = $self->{figure_of_merit}->{cum_prob_bad};
 	$o .= "$red_font_start" if $bad_FOM;
-	$o .= "Probability of acquiring 2,3, and 4 or fewer stars (10^x):\t";
-        # override formatting to match pre-JSON strings
-        foreach (2..4) {
-            $o .= substr(sprintf("%.4f", "$probs[$_]"), 0, 6) . "\t";
-        }
+	$o .= "Probability of acquiring 2 or fewer stars (10^x):\t";
```
taldcroft: I think we should be printing out `P2`, i.e. `-log10(..)`.  So this would change to `(10^-x)`.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -2656,14 +2634,235 @@ sub set_ccd_temps{
...
+        $sid = int($sid);
+        # While assigning ACQ stars into a list, clip to allowed ranges and warn as needed.
+        if ($cat_cmd->{"TYPE$i"} =~ /BOT|ACQ/){
+            push @acq_ids, $sid;;
+            my $hw = $cat_cmd->{"HALFW$i"};
+            if (($hw > 180) or ($hw < 60)){
+                push @{$self->{yellow_warn}}, sprintf(
+                    ">> WARNING: [%2d] Halfwidth %d outside range 60 to 180. Clipped for probs.",
+                    $i, $hw);
+                # Clip hw if outside the range 60 to 180 for probabilities
```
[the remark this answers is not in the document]
jeanconn: Can you restate?  I'm lost.  As far as I can tell, proseco UserWarns on all the clips it does, and our previously discussion was that we didn't want those as UserWarnings at console.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: OK, and yeah, I'd just heard not calculating probabilities for the replan/reopen observations and ignoring their in-the-starcheck-output warnings, I hadn't heard starting the starcheck output at the beginning of the DOT.  That would be more a more complicated change.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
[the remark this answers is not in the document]
taldcroft: Where do you see that?  I actually tested this prior to making the comment and didn't see warnings.
```
In [7]: aca = get_aca_catalog(21277, include_ids_acq=[1152523976], include_halfws_acq=[240])

In [8]: aca
```
Here's the line in question, which just says to extrapolate:

https://github.com/sot/proseco/blob/24ad845442861440122af028ebffbcb4ca01dfcb/proseco/acq.py#L473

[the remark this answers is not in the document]
jeanconn: Gotcha.  I didn't realize proseco was interpolating to effect a clip before passing to the grid acq model.  Doesn't matter *that* much because I think we should have visibility in starcheck if a box size outside the range used by the model or proseco is picked.  Not sure about the best way to implement in that case.

Specifically, I'm happier to also get rid of the "search box smaller that slew error" warning if I know that there is something that will caution it if we have a zero halfwidth acq box.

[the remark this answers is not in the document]
jeanconn: Oh wait, the proseco one is extrapolating?  I look at this some more.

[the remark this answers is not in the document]
taldcroft: "Extrapolating" in this case only means that it doesn't throw an exception if you give it a value outside the range. The `kind='nearest'` arg is key here, which is nearest-neighbor interpolation, or the relevant edge value for x-values outside the bounds.  It works, I tested it.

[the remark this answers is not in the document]
taldcroft: The warning on search box sizes outside of the proseco range is good, I was not implying to drop that.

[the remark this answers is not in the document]
taldcroft: And now I realize where you were seeing the UserWarning, which is that starcheck 12.1 calls the chandra_aca star probabilities model directly and can feed out-of-bounds halfwidths (e.g. JAN0419A).  But starcheck 13.0+ will use proseco, which does the clipping without issuing a warning.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -2656,14 +2634,195 @@ sub set_ccd_temps{
...
+        q1 => 0 + $targ_cmd->{q1}, q2 => 0 + $targ_cmd->{q2}, q3 => 0 + $targ_cmd->{q3}, q4 =>0 + $targ_cmd->{q4},
+        man_angle => $man_ang,
+        detector => $si, offset=> 0 + $offset,
+        dither_acq_y => $self->{dither_acq}->{ampl_y},
+        dither_acq_z => $self->{dither_acq}->{ampl_p},
+        dither_guide_y => $self->{dither_guide}->{ampl_y},
+        dither_guide_z => $self->{dither_guide}->{ampl_p},
+        t_ccd_acq => $self->{ccd_temp_acq},
+        t_ccd_guide => $self->{ccd_temp},
+        acqs => \@acq_ids,
```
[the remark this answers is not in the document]
jeanconn: acq_ids and acq_indexes is not completely unambiguous, but...

jeanconn:
I think this is good to go.  The last commit, 945fa19, is the change since the meeting today.

taldcroft:
I checked 4e289bf and the rest by running `make test` with a temporarily edited version with `P2_red=5 or 6`.  This generated the expected yellow and red warnings.

BTW `make test` seems to require that `ROOT_FLIGHT` is defined outside the test script.

PR #276, opened 2019-01-12
Use proseco for a quasi-independent hot pix check

### LINKED ISSUE #244: Report bad pixel offset with sign
The offset to a nearby bad pixel is reported as the absolute value, but knowing the sign (and the difference convention) would be useful for pre-review.

https://github.com/sot/starcheck/blob/fa0eaad500eef13f13c611d3e9bddd5504f9f7e8/starcheck/src/lib/Ska/Starcheck/Obsid.pm#L1398

jeanconn:
Use proseco for a quasi-independent hot pix check

This feeds the stars and mags from the starcheck-determined star catalog into a small routine that calls proseco.guide.get_imposter_mags and determines the worst-case centroid offset for the
candidate star and the brightest 2x2 imposter.  Those offsets are then checked against tolerances and warnings are output (> 2.5 arcsec centroid offset gives orange warn, > 4.0 offset gives red/critical warn).

Closes #244


starcheck/data/ACABadPixels
```diff
@@ -1,255 +1,3 @@
-Bad Pixel List from 2017272 2018002 2018100 Dark Currents
-
 Star.Body.Pixels.BadPixels = [ ...
- -512,  511,   -1,    0; ...
-   -1,    0, -512,  511; ...
  -245,    0,  454,  454; ...
```
jeanconn: The intent here is that only the large CCD trap remains (not sure if the fid defect should be added).

Functionally I just checked that an observation that previously had a star with a trap-related warning still did so with this bad pixel list.  Obsid 48812 in OCT2918A
```
>> WARNING: [ 8] Nearby ACA bad pixel.  row, col (-64, 454), dy, dz (0, 6) 
```

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -23,14 +23,47 @@ use warnings;
...
 
+from proseco.core import ACABox
+from proseco.guide import get_imposter_mags
+
 def _guide_count(mags, t_ccd):
     return float(guide_count(np.array(mags), t_ccd))
 
+def check_hot_pix(idxs, yags, zags, mags, types, t_ccd, date, dither_y, dither_z, dmag=1.0):
+
+    dark = aca_dark.get_dark_cal_image(date=date.decode('ascii'),
```
jeanconn: I should probably see about getting the date of the dark cal put out in the processing summary.

jeanconn: It is also getting fetched once per obsid in this configuration, but would be a few more steps to set it and pass it per-schedule.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -23,14 +23,47 @@ use warnings;
...
+                                           t_ccd_ref=float(t_ccd), aca_image=True)
+
+    guide_stars = []
+    for i, y, z, mag, ctype in zip(idxs, yags, zags, mags, types):
+        ctype = ctype.decode('ascii')
+        if ctype == 'BOT' or ctype == 'GUI':
```
jeanconn: Should probably set this up to allow FID as well, but then call the whole routine on fids separately with some error term instead of dither.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1231,6 +1262,31 @@ sub check_star_catalog {
...
+            push @idxs, $i;
+            push @yags, $c->{"YANG$i"};
+            push @zags, $c->{"ZANG$i"};
+            # Add zero to get items that look more like float values in the arrays
+            push @mags, ($c->{"GS_MAG$i"} eq '---') ? 13.94 : $c->{"GS_MAG$i"} + 0.0;
+            push @types, $c->{"TYPE$i"};
+        }
+    }
+
+    # Run the hot pixel region check on the Python side on FID|GUI|BOT
```
jeanconn: Only running on GUI|BOT despite comment.

taldcroft:
What worries me here is that this check gives an OK to a case with a 7.0 mag star and a 8.01 mag imposter that could be at +3, +3 row/col relative to the nominal position.  So basically that would be a total fail of a guide star.  Either very large centroid offsets or pounding the background pixels.

So one can respond by making `dmag=3.0`, but then that ends up failing basically every 10.2 mag star.

I think what we run into here is the same issue that we solved for spoilers, namely one needs to look at centroid offsets that result.  Of course for an imposter this is a little trickier because (from the star readout perspective) the imposter is dithering around.

So that's where in the meeting I had the reaction "let's check whatever proseco is normally checking for imposters".  And having now just checked the code, it seems like a reasonable solution would just be re-casting this in terms of `get_pixmag_for_offset(cand_mag, offset)` and having the relevant check parameter be the allowed centroid `offset`.

This does not address the imposter getting into the background.  For now we can live with that (esp. since the proseco selection process also does not account for this issue).

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -23,14 +23,68 @@ use warnings;
...
+    :param types: catalog TYPE (ACQ|BOT|FID|MON|GUI) as list or array
+    :param date: observation date (bytestring via Inline)
+    :param dither_y: dither_y in arcsecs (guide dither)
+    :param dither_z: dither_z in arcsecs (guide dither)
+    :param dmag: delta mag for warning (default 1.0 mag)
+    :return fails: list of dictionaries with keys that define the index, the imposter mag, and star or fid
+             info to make a warning.
+    """
+
+
```
taldcroft: One blank line.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -23,14 +23,68 @@ use warnings;
...
+    dark = aca_dark.get_dark_cal_image(date=date.decode('ascii'),
```
taldcroft: I realize this ship has already sailed, but for new Inline Python code like this, how about starting a convention to do all the type-correction code in a block at the beginning.  That makes it all tidy and clean afterward (i.e. you know for the rest of the function that `dither_z` is a nice Python float).  So
```
idxs = [int(idx) for idx in idxs]
...
dither_y = float(dither_y)
ctype = ctype.decode('ascii')
```

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -23,14 +23,68 @@ use warnings;
...
+                                       t_ccd_ref=float(t_ccd), aca_image=True)
+    fails = []
+    for i, y, z, mag, ctype in zip(idxs, yags, zags, mags, types):
```
taldcroft: Use `idx`, `yag`, `zag`.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -23,14 +23,68 @@ use warnings;
...
+        ctype = ctype.decode('ascii')
+        if ctype in ['BOT', 'GUI', 'FID']:
+            if ctype in ['BOT', 'GUI']:
+                dither = ACABox((float(dither_y), float(dither_z)))
+            else:
+                dither = ACABox((5.0, 5.0))
+            row, col = yagzag_to_pixels(float(y), float(z))
+            entry = Table([{'i': int(i), 'row': row, 'col': col, 'mag': float(mag), 'type': ctype}])
```
taldcroft: Should be `entries = Table(..)` then
```
entry = entries[0]
```
The code below `float(entry['row'])` is awkward because technically you are taking `float` of an array.  Which fails if the array is not exactly 1 element long, so basically it is accidentally succeeding in a way that might change in the future.

jeanconn:
I'm a little confused though.  It sounds like you think get_pixmag_for_offset is smarter than it is, because I think basically, because it doesn't move the bad pixel around and find any kind of worst case, and just estimates that putting all the counts from the pixel on the edge of the 2x2 is the worst case, you get basically a fixed delta mag relationship of around 1 mag.

Or maybe the math for get_pixmag_for_offset is just wrong, but:
```
In [4]: proseco.guide.get_pixmag_for_offset(7, 4.5)
Out[4]: 7.9199419632364858
In [5]: proseco.guide.get_pixmag_for_offset(8, 4.5)
Out[5]: 8.9199419632364858
In [6]: proseco.guide.get_pixmag_for_offset(9, 4.5)
Out[6]: 9.9199419632364858
In [7]: proseco.guide.get_pixmag_for_offset(10, 4.5)
Out[7]: 10.919941963236486
In [8]: proseco.guide.get_pixmag_for_offset(11, 4.5)
Out[8]: 11.919941963236486
```

jeanconn:
The guide star selection basically has the "hope" that a less spoiled star than the 7.0 mag star with the 8.01 mag imposter in your scenario is selected, but does not really have explicit protection against it.

taldcroft:
So I didn't appreciate that the last-stage centroid offset is so large, 4.5 arcsec.  I would consider that entirely unacceptable for a guide star.  Your code examples bear this out - I had said a delta mag of 1.0 is no good, and now you say that a delta mag of 1.0 corresponds to a centroid offset that is likewise no good.  So we're on the same page.

But it has led me to an idea for hot-pixel checking that might work.  Well at least it makes sense to me at this moment: for the brightest identified imposter, compute the max centroid offset assuming the imposter is at the edge (same as `get_pixmag_for_offset`).  Take the max of the y and z offsets.  Then:
- Critical warning for offset > 4.0 arcsec (basically useless guide star)
- Else Red warning for offset > 2.5 arcsec (would rather not use it)

So with this, the checking is keyed off a physically meaningful test (even though we accept that the computation of the offset has limitations).  The test leverages proseco code but has independent limits.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: Even better if we could just de_bytestr() everything to start, but I haven't figured out the best way to just stick that in here.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: For myself, the single entry table made more sense as an `entry` but OK.  get_imposter_mags needs a table not a row though, so it is mix and match.

taldcroft:
I haven't looked at recent stats, but I think that residuals above or near 5 arcsec *are* unusual, which is good because the OBC will be tossing the star at that point.  Residuals above 3 arcsec are not unusual.

About warning on stage 5, not so keen.  I'd much prefer checks that relate to some actual performance metric if possible.  In the present implementation, that just brings in either a bright imposter (which will be warned) or a color=0.7 guide star (which will be warned).  BTW, `allow_0p7_color_star` would be more informative than `DoBminusVcheck` and would have saved me having to check the code :smile:.

However, this does point to a feature that I'm not sure we have discussed, which is putting the guide star stage into the starcheck report in the usual place.  Not a requirement for 13.0, but it would good at some point.  Also, it might be illuminating to add a histogram of guide star stage into the proseco guide 4.3.2 validation report.  Have you ever looked at anything like that?

jeanconn:
Regarding the stage selected, I think I looked a bit, but still intended to do more (with more stage parameter review and optimization).

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: With regard to getting the date in the processing summary, @taldcroft , do you want me to head in that direction or leave this alone?   It seems like the implementation would be to fetch the dark cal in starcheck.pl and try to either set as a Python global or pass to Obsid object or check_starcheck_catalog.

Could also do something like use one date for all these calls and then *infer* the dark cal use in starcheck.pl.

taldcroft: I think this can wait for a subsequent release, so I recommend to just open an issue.  We might help ourselves out by taking milestoning more seriously, so make a new milestone and tag thusly.  This can help separate things we actually want for a release from, well, stuff that comes up.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -23,14 +23,97 @@ use warnings;
...
+                # a single-candidate table `entries`
+                entries = Table([{'idx': idx, 'row': row, 'col': col, 'mag': mag, 'type': ctype}])
+                imp_mags, imp_rows, imp_cols = get_imposter_mags(entries, dark, dither)
+                offset = imposter_offset(mag, imp_mags[0])
+                imposters.append({'idx': int(idx), 'status': int(0),
+                                  'entry_row': float(row), 'entry_col': float(col),
+                                  'bad2_row': float(imp_rows[0]), 'bad2_col': float(imp_cols[0]),
+                                  'bad2_mag': float(imp_mags[0]), 'offset': float(offset)})
+            except:
+                imposters.append({'idx': int(idx), 'status': int(1)})
```
jeanconn: I don't want to necessarily try to skip running this code for a star or fid off the ccd or too close to the edge, so having some error handling seems to make sense.  Whether that handling should just skip the questionable idx altogether, provide a more specific error message, or do what I have here which is to pass back just a failed status, is a separate question.  Thus far, this has only failed for undercovers with fid lights commanded off the ccd in regression weeks.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1239,6 +1320,44 @@ sub check_star_catalog {
...
+        }
+        my $warn = sprintf("$alarm [%2d] Imposter mag %.1f offset %.1f (row % 4d, col % 4d) star (% 4d, % 4d)\n",
+                           $imposter->{idx}, $imposter->{bad2_mag}, $imposter->{offset},
+                           $imposter->{bad2_row}, $imposter->{bad2_col},
+                           $imposter->{entry_row}, $imposter->{entry_col});
+        if ($imposter->{offset} > 4.0){
+            push @warn, $warn;
+        }
+        elsif ($imposter->{offset} > 2.5){
+            push @orange_warn, $warn;
```
jeanconn: @taldcroft Your suggestion was critical warn on > 4 and red warn on > 2.5, but critical and red are the same, so I went with orange.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: Sounds good.  Another advantage of calling get_dark_cal_image right here is that we're getting the temperature scaling per obsid without needing to do anything more manually.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
taldcroft: Failed status like you have seems fine.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -23,14 +23,97 @@ use warnings;
...
+    :param dither_y: dither_y in arcsecs (guide dither)
+    :param dither_z: dither_z in arcsecs (guide dither)
+    :param yellow_lim: yellow limit centroid offset threshold limit (in arcsecs)
+    :param red_lim: red limit centroid offset threshold limit (in arcsecs)
+    :return imposters: list of dictionaries with keys that define the index, the imposter mag,
+             a 'status' key that has value 0 if the code to get the imposter mag ran successfully,
+             calculated centroid offset, and star or fid info to make a warning.
+    """
+
+    idxs = [int(idx) for idx in idxs]
```
taldcroft: I've forgotten why all this input casting is required.  I understand the unicode thing, and casting from np array to pure Python types on output, but why the `float` / `int` casting?

taldcroft: > on input I think the weak Perl types have caught me up

That's a fair enough answer (i.e. Perl idiotically passing `"1.2"` instead of a float).  Every time I see `$value + 0.0` it makes my stomach turn but that's what it is.

jeanconn: I cut them anyway in a new commit at the end of this PR and did enough of the regression tests that I think this is fine.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -65,16 +65,15 @@ def check_hot_pix(idxs, yags, zags, mags, types, t_ccd, date, dither_y, dither_z
              a 'status' key that has value 0 if the code to get the imposter mag ran successfully,
              calculated centroid offset, and star or fid info to make a warning.
     """
-
-    idxs = [int(idx) for idx in idxs]
-    yags = [float(yag) for yag in yags]
-    zags = [float(zag) for zag in zags]
-    mags = [float(mag) for mag in mags]
+    idxs = [idx for idx in idxs]
```
taldcroft: This should be written as `list(idxs)` if that is indeed necessary.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -65,16 +65,15 @@ def check_hot_pix(idxs, yags, zags, mags, types, t_ccd, date, dither_y, dither_z
...
+    yags = [yag for yag in yags]
+    zags = [zag for zag in zags]
+    mags = [mag for mag in mags]
     types = [t.decode('ascii') for t in types]
-    t_ccd = float(t_ccd)
+    t_ccd = t_ccd
     date = date.decode('ascii')
-    dither_y = float(dither_y)
-    dither_z = float(dither_z)
+    dither_y = dither_y
```
taldcroft: Time to slow down, maybe take a break?  :smile:

jeanconn: Hah.  I suppose that doesn't quite satisfy the intent to remove unnecessary code...

jeanconn: Everything seems fine in limited testing with more aggressive cutting of this conversion stuff, so I think it is fine now.

starcheck/data/ACABadPixels
jeanconn: I did not add a comment in the file about the trap as I didn't want to play games with the parser.

PR #277, opened 2019-01-12
More small dither-related changes

jeanconn:
More small dither-related changes

This PR

1) puts the tiny bugfix to short-circuit dither checks sooner (when no observation start time available to do them). 
2) tracks the maximum guide amplitude in y and z over the observation instead of just the guide amplitudes 8 minutes after maneuver end (should only matter for big dither but...)
3) adds a green warning for observations that have different acq and guide dither (again should only be interesting for big dither)

### DIFF
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -720,6 +720,8 @@ sub check_dither {
 
     $self->{dither_acq} = $acq_dither;
     $self->{dither_guide} = $guide_dither;
+    $self->{dither_guide_y_max} = $guide_dither->{ampl_y};
+    $self->{dither_guide_z_max} = $guide_dither->{ampl_p};
 
     # Check for standard dither
     if ($guide_dither->{state} eq 'ENAB'){
@@ -754,6 +756,10 @@ sub check_dither {
     }
     else{
         foreach my $dither (reverse @{$dthr}) {
+            if ($dither->{time} < $obs_tstop){
+                $self->{dither_guide_z_max} = max(($dither->{ampl_p}, $self->{dither_guide_z_max}));
+                $self->{dither_guide_y_max} = max(($dither->{ampl_y}, $self->{dither_guide_y_max}));
+            }
             if ($dither->{time} > ($obs_tstart + $obs_beg_pad)
                     && $dither->{time} <= $obs_tstop - $obs_end_pad) {
                 push @{$self->{warn}}, "$alarm Dither commanding at $dither->{time}.  During observation.\n";
@@ -764,8 +770,11 @@ sub check_dither {
         }
     }
 
-
-
+    if (($self->{dither_guide_y_max} != $self->{dither_guide}->{ampl_y})
+            or ($self->{dither_guide_z_max} != $self->{dither_guide}->{ampl_p})){
+        push @{$self->{yellow_warn}}, sprintf("$alarm Max Y Z ampl during guide used for checking Y=%.1f Z=%.1f \n",
+                                              $self->{dither_guide_y_max} + 0.0, $self->{dither_guide_z_max} + 0.0);
+    }
 
     # For eng obs, don't have OR to specify dither, so stop before doing vs-OR comparisons
     if ( $self->{obsid} =~ /^\d*$/){
@@ -1172,9 +1181,9 @@ sub check_star_catalog {
 	$dither_acq_z = 20.0;
     }
 
-    if (defined $self->{dither_guide}){
-	$dither_guide_y = $self->{dither_guide}->{ampl_y};
-        $dither_guide_z = $self->{dither_guide}->{ampl_p};
+    if (defined $self->{dither_guide}->{ampl_y_max}){
+	$dither_guide_y = $self->{dither_guide}->{ampl_y_max};
+        $dither_guide_z = $self->{dither_guide}->{ampl_p_max};
     } else {
 	$dither_guide_y = 20.0;
 	$dither_guide_z = 20.0;
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -1933,7 +1933,9 @@ sub print_report {
     if ( ( defined $self->{ra} ) and (defined $self->{dec}) and (defined $self->{roll})){
 	$o .= sprintf "RA, Dec, Roll (deg): %12.6f %12.6f %12.6f\n", $self->{ra}, $self->{dec}, $self->{roll};
     }
-    if (defined $self->{dither_guide}){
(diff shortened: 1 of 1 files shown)


jeanconn:
For obsid 20168 this basically looks like this at the top
```
Dither:  ON Y_amp=64.0  Z_amp= 8.0  Y_period=2647.1  Z_period=1000.0 
```
And this down in the warnings
```
>> WARNING: Non-standard dither
>> WARNING: Reviewed with ACQ dither Y=20.0 Z=20.0 
>> WARNING: Max Y Z ampl during guide used for checking Y=64.0 Z=20.0 
>> INFO   : Observation passes 'big dither' checks
```

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -707,17 +712,18 @@ sub check_dither {
 
     $self->{dither_acq} = $acq_dither;
     $self->{dither_guide} = $guide_dither;
+    $self->{dither_guide_y_max} = $guide_dither->{ampl_y};
```
taldcroft: Seems like this would be better as an attribute of `$self->{dither_guide}`, i.e. `$self->{dither_guide}->{ampl_y_max}`.  It doesn't really belong as a top-level attribute of the Obsid class.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -707,17 +712,18 @@ sub check_dither {
...
+    $self->{dither_guide_z_max} = $guide_dither->{ampl_p};
```
taldcroft: Not sure if this is opening a can of worms, but you should be consistent and use `p(itch) / y(aw)` everywhere or `y(angle) / z(angle)` everywhere.  Otherwise it gets very confusing.  Since existing code seems to be pitch/yaw then that might be the easiest choice.  I don't have a strong preference except to use one system.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -707,17 +712,18 @@ sub check_dither {
...
             push @{$self->{yellow_warn}}, "$alarm Non-standard dither\n";
         }
-        else{
-            if (($guide_dither->{ampl_p} != $acq_dither->{ampl_p})
-                    or ($guide_dither->{ampl_y} != $acq_dither->{ampl_y})){
-                push @{$self->{yellow_warn}}, "$alarm Dither changes between ACQ and GUI \n";
-            }
+        if (($guide_dither->{ampl_p} != $acq_dither->{ampl_p})
+                or ($guide_dither->{ampl_y} != $acq_dither->{ampl_y})){
+            push @{$self->{yellow_warn}}, sprintf("$alarm Reviewed with ACQ dither Y=%.1f Z=%.1f \n",
```
taldcroft: The seems like an Info message to me.  There is nothing wrong or actionable here, just giving the reviewer a heads-up of something slightly unusual.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -751,8 +761,11 @@ sub check_dither {
         }
     }
 
-
-
+    if (($self->{dither_guide_y_max} != $self->{dither_guide}->{ampl_y})
+            or ($self->{dither_guide_z_max} != $self->{dither_guide}->{ampl_p})){
+        push @{$self->{yellow_warn}}, sprintf("$alarm Max Y Z ampl during guide used for checking Y=%.1f Z=%.1f \n",
```
taldcroft: This seems like Info to me.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1159,9 +1172,9 @@ sub check_star_catalog {
 	$dither_acq_z = 20.0;
     }
 
-    if (defined $self->{dither_guide}){
-	$dither_guide_y = $self->{dither_guide}->{ampl_y};
-        $dither_guide_z = $self->{dither_guide}->{ampl_p};
+    if (defined $self->{dither_guide_y_max}){
+	$dither_guide_y = $self->{dither_guide_y_max};
+        $dither_guide_z = $self->{dither_guide_z_max}
     } else {
```
taldcroft: If I understand the approach now, this case (dither amplitude unknown) should issue a yellow warning since it means something unexpected has happened.  Basically we *always* expect to have valid dither amplitudes for every observation, right?

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -686,6 +686,11 @@ sub check_dither {
     my $obs_tstart = $self->{obs_tstart};
     my $obs_tstop = $self->{obs_tstop};
 
+    unless (defined $obs_tstart){
+        push @{$self->{warn}}, "$alarm Cannot determine obs start time for dither. \n";
```
taldcroft: Add in `Cannot determine obs start time for dither, not checking.`

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: Seems file.  I actually had something more like that, ran into some Perl issues and then ended up with simpler data structure.  Found out 2 hours later my issue had been an extra $, so I think I can manage this time around...

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: There's the question on whether a real "heads-up" should be yellow.  I was mostly thinking that the action is that should get some extra attention and the question is if that extra attention will happen without with a yellow warning count in the summary.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: I think "everywhere" is a can of worms, but I'll at least try for check_dither.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: That seems right to me (red or orange would actually seem more appropriate to me at this point).  With regard to kadi dynamic states integration I still have questions (for example, on historical test schedules which may or may not have initial conditions that match kadi?) but should be OK for dither.

jeanconn:
I didn't go crazy convertion y/z to y/p but made some progress on that front and think I've addressed the rest of the comments.  I've only test run this against JAN0419T and AUG2018T thus far (was planning to run and review more regression tests outputs on the full 13.0 candidate).

jeanconn:
Is this GTG or do I need to try to remove each line to see if it is no longer necessary?

jeanconn:
After reviewing some regression test outputs, I moved the new dither checking at the top check_star_catalog after the check to see if there is actually a star catalog (even though this was already approved).  Rerunning regression outputs, but I think this change is non-controversial and will plan to merge after double-checking outputs.

PR #280, opened 2019-01-15
Check to see if obsid has catalog before giving up for RDE obsids

jeanconn:
Check to see if obsid has catalog before giving up trying to set dither for RDE obsids.

### DIFF
--- a/starcheck/data/characteristics.yaml
+++ b/starcheck/data/characteristics.yaml
@@ -55,6 +55,8 @@ paths:
    week_lookup: https://icxc.harvard.edu/cgi-bin/aspect/starcheck/find_shortterm.cgi
 
 # string match on oflsids that may not have a maneuver or starcat, and its OK!
+# The RDE (and probably RDX) entries most likely do not applie to ORViewer DOT schedules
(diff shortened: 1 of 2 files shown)


jeanconn:
It looks like there are ERs with RDE in the oflsid that have star catalogs and there are some that do not, and previous logic to handle SIR RDE oflsids meant they were skipped in the dither checking.  This didn't really matter when all ERs were skipped wrt to dither (and assigned dither amp = 20), but now after #266 when we're trying to get and assign a reasonable dither for all observations, this is more of an issue.

taldcroft:
I'm lost.  I thought RDE/RDX is no longer a concept with ORviewer DOT generation.  I grepped for RDE, RDX, and DC_T in all the 2018 DOTs and none of them appear.  Dark cal observations appear like `O48544M002`.

jeanconn:
Though I honestly thought that this was a noncontroversial edit (check to see if there is a star catalog before deciding to skip with regard to assigning dither) that just works.

If you want a more comprehensive starcheck cleanup, that's also a different issue.

PR #281, opened 2019-01-15
Do not check observations prior to schedule start (DOT)

### LINKED ISSUE #101: Handle Replan/Reopen schedules correctly

taldcroft:
This skips starcheck checking of backstop commands which occur prior to the formal schedule start.  

This applies to replan/reopen products where backstop commanding from a prior schedule is included in the command loads but was not generated in the schedule.  The implementation relies on the DOT to define commands which are subject to starcheck review.

See the icxc aspect area tmp/jan0419a.html for an example output.  The global command warnings remain, along with the dynamic offsets errors due to obsids that are missing from the OR list.

Closes #101

### DIFF
--- a/starcheck/src/starcheck.pl
+++ b/starcheck/src/starcheck.pl
@@ -256,7 +256,6 @@ for my $data_file ('up.gif', 'down.gif', 'overlib.js'){
 
 
 # First read the Backstop file, and split into components
-my $bogus_obsid = 1;
 my @bs = Ska::Parse_CM_File::backstop($backstop);
 
 my $i = 0;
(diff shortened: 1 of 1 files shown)


taldcroft:
> Also, do you want me to remove an changes in #272 that are unnecessary combined with this change?

Yes.  Probably makes sense for you to review / test this PR until you are satisfied, merge, then rebase #272 and remove changes in #272 that become unnecessary.

jeanconn:
I'm still not sure if I should find any really weird things to throw at this, but it looks fine to me in regression testing and the code changes, so moving forward.

jeanconn:
Though as you noted at top, the warnings on the dynamic aimpoint checks remain which, from a public perspective, does mean that starcheck's output isn't *strictly* starting at schedule start.

taldcroft:
This is carried over to #287 now for the dynamic aimpoints checking.

PR #290, opened 2019-01-18
Add manual bad star 1158290496 (bad position / PM)

taldcroft:
In obsid 19763 this was seen with a centroid offset of around 5 arcsec, generating a worst-case situation where it is persistently just inside the Kalman limit threshold (5 arcsec box).  It may be that the PM is wrong (doesn't exist) or the position is wrong, but in any case we should not use this star.

I think we can merge this with only a "smoke" test without re-doing the entire regression suite.

### DIFF
--- a/starcheck/data/agasc.bad
+++ b/starcheck/data/agasc.bad
@@ -1,3 +1,5 @@
+1158290496 Incorrect position or proper motion (obsid 19763) (TA 18-Jan-2019)
+350392600 Too bright, 3.6 mag in obsid 21252, catalog mag 7.6 (TA/JMC 17-Jan-2019)
 797847184 not found as GS in 49383 or 49382, exclude per Tom 12-Mar-2018
 914493824 Too bright, exclude per Eric 15-Aug-2016
 509225640 Too bright, 5.1 mag, exclude per Eric (JMC 17-Feb-2015)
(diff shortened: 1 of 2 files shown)


taldcroft:
Re-running AUG1707 would be a solid test.

taldcroft:
I checked by grepping that 1158290496 is not in JAN1419 nor JAN2119.

jeanconn:
Though updating the manual_bad_star list doesn't actually effect the change.  I suppose for #288 I should have made the edit to agasc.bad as well, so I'll add that.

jeanconn:
The intent was that one would update manual_bad_stars, and then run bad_stars.py to make a new agasc.bad with manual stars and automatic additions from the database.  Since we haven't stabilized again on the right filter for automatic additions, will just update agasc.bad directly.

jeanconn:
Or we could change the code to read from the manual bad star list and a separate automatic list.  Whatever.

taldcroft:
All this needs to be fully overhauled with a Ska-wide approach.  

Local to starcheck I'm confused about how `bad_gui_stars.rdb` fits in the picture, and why it appears the last run of `bad_stars.py` for auto-bad-stars was 4 years ago.

taldcroft:
In any case you should just do what is needed to get this bad star into starcheck 13.0.  This is critical because I believe using it is likely to lead to a safing action.

jeanconn:
Not sure why you are confused that the last run of bad_star.py was 4 years ago.   Time does fly and I think we went back and forth on criteria a bunch of times and then didn't do anything.

With regard to bad_gui_stars.rdb, somewhat the same deal, only those are the starcheck-local database files that aren't used for flight starcheck (only used if no access to the database).

PR #294, opened 2019-01-22
Remove extra space in Dither line of report

jeanconn:
Remove extra space in Dither line of report.


jeanconn:
The introduced space looks to break parsing of the dither line in mica.starcheck.starcheck_parser and does not seem to greatly improve readability.  Looks to me like it can go.

jeanconn:
Space looks introduced as side-effect of a141c68

PR #302, opened 2019-03-19
Updates for osx compat

### LINKED ISSUE #263: Doesn't quite work on osx
Looks like there are still a few tweaks that would be needed to run on OSX.

The starcheck  wrapper is calling /bin/mktemp which doesn't exist.  Maybe just mktemp would be OK.

The starcheck wrapper expects SKA_ARCH_OS to be defined.  If we want to reduce Ska3 dependence to SKA, this should just be set to SKA.

The Dark Cal checker expects SKA_SHARE and SKA_DATA to be defined.  These should probably be set to be their SKA equivalents or just reference the package area appropriately.

The initial conditions for the temperature model could probably just be set with maude to avoid the eng_archive access dependency.

jeanconn:
Updates for osx compat

Closes #263

### DIFF
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -100,7 +100,7 @@ def get_options():
 
 def get_ccd_temps(oflsdir, outdir='out',
                   json_obsids=sys.stdin,
-                  model_spec=None, char_file=None, orlist=None, use_maude=0,
+                  model_spec=None, char_file=None, orlist=None,
                   verbose=1, **kwargs):
     """
     Using the cmds and cmd_states sybase databases, available telemetry, and
@@ -112,30 +112,23 @@ def get_ccd_temps(oflsdir, outdir='out',
     :param json_obsids: file-like object or string containing JSON of
                         starcheck Obsid objects
     :param model_spec: xija ACA model specification
-    :param use_maude: if 1, use maude to fetch recent AACCCDPT data
     :param verbose: Verbosity (0=quiet, 1=normal, 2=debug)
     :returns: JSON dictionary of labeled dwell intervals with max temperatures
     """
     if not os.path.exists(outdir):
         os.mkdir(outdir)
 
-    if use_maude == 1:
-        fetch.data_source.set('maude allow_subset=False')
-
     config_logging(outdir, verbose)
 
     # Store info relevant to processing for use in outputs
     proc = {'run_user': os.environ['USER'],
             'run_time': time.ctime(),
-            'errors': [],
-            'use_maude': use_maude,
-            }
+            'errors': []}
     logger.info('##############################'
                 '#######################################')
     logger.info('# %s run at %s by %s'
                 % (TASK_NAME, proc['run_time'], proc['run_user']))
     logger.info('# {} version = {}'.format(TASK_NAME, VERSION))
-    logger.info('# Fetch data sources set to {}'.format(fetch.data_source.sources()))
     logger.info('###############################'
                 '######################################\n')
 
@@ -160,10 +153,10 @@ def get_ccd_temps(oflsdir, outdir='out',
     proc['datestart'] = DateTime(tstart).date
     proc['datestop'] = DateTime(tstop).date
 
-    # Get temperature telemetry for 7 days prior to min(tstart, NOW)
+    # Get temperature telemetry for 30 days prior to min(tstart, NOW)
     tlm = get_telem_values(min(tstart, tnow),
-                           ['aacccdpt'],
-                           days=7)
+                           ['aacccdpt', 'pitch'],
+                           days=30)
 
     states = get_week_states(tstart, tstop, bs_cmds, tlm)
     # if the last obsid interval extends over the end of states
@@ -375,7 +368,8 @@ def mock_telem_predict(states):
 
     tlm = fetch.MSIDset(['aacccdpt'],
                         states[0]['tstart'],
-                        states[-1]['tstart'])
+                        states[-1]['tstart'],
+                        stat='5min')
     temps = {'aca': tlm['aacccdpt'].vals}
     return tlm['aacccdpt'].times, tlm['aacccdpt'].vals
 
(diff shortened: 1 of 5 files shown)


jeanconn:
I'd have liked to toss in the use of maude to make this more osx-mobile-ready, but the ACA model code  is set to fetch 30 days of AACCCDPT, so for operational use I don't think we want to swap that over to use the full resolution data from the 5min stat it is doing, and it looks like we don't have stat support in fetch just yet.

taldcroft:
`AACCCDPT` is sampled at 32.8 sec, so a MAUDE or CXC fetch of 30 days full resolution is no problem.  About 80000 samples, which I checked is done in a couple of seconds.  So got for it!

jeanconn:
A MAUDE fetch of 30 days full resolution doesn't immediately seem to be option either:

Exception: MAUDE query failed: cannot query more than 7 days with allow_subset=False at line 28

starcheck/calc_ccd_temps.py
```diff
@@ -404,7 +410,7 @@ def get_telem_values(tstart, msids, days=7, name_map={}):
     start = DateTime(tstart - days * 86400).date
     stop = DateTime(tstart).date
     logger.info('Fetching telemetry between %s and %s' % (start, stop))
-    msidset = fetch.MSIDset(msids, start, stop, stat='5min')
+    msidset = fetch.MSIDset(msids, start, stop)
```
jeanconn: I think it is still fine to interpolate in 5 minute steps below, but if we are just fetching the one msid the interpolation doesn't seem necessary anyway.

starcheck/calc_ccd_temps.py
```diff
@@ -153,10 +160,10 @@ def get_ccd_temps(oflsdir, outdir='out',
     proc['datestart'] = DateTime(tstart).date
     proc['datestop'] = DateTime(tstop).date
 
-    # Get temperature telemetry for 30 days prior to min(tstart, NOW)
+    # Get temperature telemetry for 7 days prior to min(tstart, NOW)
     tlm = get_telem_values(min(tstart, tnow),
-                           ['aacccdpt', 'pitch'],
-                           days=30)
+                           ['aacccdpt'],
```
jeanconn: I cut the pitch fetching because it both would not really work on maude and because I don't recall why it was being fetched in the first place.

starcheck/calc_ccd_temps.py
```diff
@@ -153,10 +160,10 @@ def get_ccd_temps(oflsdir, outdir='out',
...
+                           days=7)
```
jeanconn: What needs to be verified to go with this 7 day fetch instead of 30?

starcheck/calc_ccd_temps.py
jeanconn: Is the interpolation necessary or can it be cut?

starcheck/calc_ccd_temps.py
taldcroft: Not sure either.  Pitch is obviously a requirement for computing the model, but that comes from the states here (if I understand).

starcheck/calc_ccd_temps.py
taldcroft: I don't know.  I haven't dug into the code to see how it gets used, so your call.

jeanconn:
OK.  I read your "So go for it!" feedback as sort-of direction to go this way, but can back it out.

I also didn't see the starcheck code as a real general example? and thought there might be some benefit to the maude strategy as being both more up-to-date and a great fallback plan.

jeanconn:
Also, what's the status of ska_sync and eng_archive msids?

taldcroft:
What do I need to do for testing?  I did:
- Copy JUN1019 directory to a local dir
- Update my ska_sync to include eng_archive files per https://github.com/sot/ska_testr/pull/15#issuecomment-499724331
- `conda install -c $ska3conda ska3-perl`

This gave me:
```
(ska3) monkeys$ ./sandbox_starcheck -dir ~/ska/data/mpcrit1/mplogs/2019/JUN1019 -out test
Fatal Python error: Py_Initialize: unable to load the file system codec
ModuleNotFoundError: No module named 'encodings'

Current thread 0x00007fff9841b380 (most recent call first):
Failed to autogenerate /private/var/folders/k1/2ktyqf0x2n7_fzvdnvljq2240000gq/T/starcheck_inline.XXXXXX.wUsWJXGx/config-darwin-2level-5.026002.

 at starcheck/src/lib/Ska/Starcheck/Obsid.pm line 162.
BEGIN failed--compilation aborted at starcheck/src/lib/Ska/Starcheck/Obsid.pm line 162.
Compilation failed in require at ./starcheck/src/starcheck.pl line 33.
BEGIN failed--compilation aborted at ./starcheck/src/starcheck.pl line 33.
```

jeanconn:
Did you source into ska?

taldcroft:
OK, success, wow!!

I encountered some difficulties because sourcing ska_envs redefines $SKA to a new location which is not my SKA.  So I made a data soft link but then made the mistake of running ska_sync with the new SKA, which blew away the soft link and made a second fresh data repo.  So I would say this is a little fiddly right now for standalone Ska, but it does work!

jeanconn:
Maybe open a skare3 issue on the problem you had with ska_envs redefining SKA?  I thought I had figured out the non-root environment scenarios but maybe missed something.

Starcheck might, in fact, work with just SKA set (without setting LD_LIBRARY_PATH and such from ska_envs.sh) but that wasn't how I was testing.

With regard to regression, I can't see how the changes that ended up in this PR could introduce regressions in flight starcheck on linux.  Do you also want to certify osx for load review?  I wasn't sure what we needed to test or if "it looks like it works" might be just fine for now.

taldcroft:
If I use `export PYTHONHOME=$CONDA_PREFIX` instead of `$SKA` then both `starcheck/src/starcheck` and `sandbox_starcheck` work in a standalone env.

taldcroft:
About regressions, I was not implying that the linux output could change, I was wondering if the OSX about matches the linux output.  That is likely the case, but does need checking for at least one load (by diffing the starcheck.txt output).

I don't foresee certifying use on OSX for flight load review right away, but I will certainly use starcheck on my mac and it would be a problem if it were giving results inconsistent with linux.

jeanconn:
OK, so  was the regression comment an aside, do you want to see regression validation for osx in this PR, or should it be a new issue?

jeanconn:
I think the real problem is that I replace SKA_ARCH_OS with SKA in these scripts, but that we've decided that SKA only defines the path of data.  For interaction with the Perl I think I do need something that always points to the Python installation (equivalent to old SKA_ROOT?), and that's a little counter to:

https://github.com/sot/skare3/issues/22#issuecomment-412646934

I think CONDA_PREFIX only works in an env, so would you want to do something like try it first?  My thought was that we just want ska_envs.sh to try to do the right thing if possible and perhaps set SKA and SKA_ROOT to reasonable things in the env.

taldcroft:
For the record, I confirmed that the report `.txt` file on my Mac for JUN1019A is exactly the same as the output from linux except for the header info up top (including the sybase error on OSX).  In both cases I ran ./starcheck_sandbox from the git repo.

Issue #246, opened 2018-06-08
Downgrade ACQ spoiler and search box size from red to orange


taldcroft:
Seems like these are not really "critical" since they would just result in loss of one acq star (worst case):
- WARNING: [ 5] Search Box Size. Star too faint for > 120 box. 
- WARNING: [ 3] Search spoiler.  738856128: Y,Z,Radial,Mag seps:  86 131 157 -0.1

jeanconn:
OK.  I had the spoiler one as #236 but didn't have the search box one.

taldcroft:
This came up again in MAR0419 with obsids 48343 and 48342.  Both of those critical warns should be orange.  

Eventually when we have enough flight experience with proseco then we may downgrade further.  Basically proseco has a nuanced way of dealing with spoilers and human intuition may not be helpful.  I suspect that spoiler messages will cease to be actionable.

jeanconn:
Thought I also don't think starcheck separates out spoiler handling for acq and guide much, so probably will need to do that too.

jeanconn:
I assume for this issue you are mostly talking about downgrading ACQ warnings only, right @taldcroft ?

taldcroft:
Yes I am just thinking of acq warnings.

jeanconn:
For the search box one, I suggest just cutting the warning altogether in #300 .

jeanconn:
Search box size done in #300 .  Spoiler warning is in #236 so closing this issue.

PR #303, opened 2019-03-19
Reduce acq spoiler from red to orange and add new red warn for guide spoiler

### LINKED ISSUE #236: Red search spoiler warning on acq star should be orange warning
Red search spoiler warning on acq star should be orange warning.

I think a single red search spoiler warning on an acq star in a catalog is not Critical and is only an issue if the overall acquisition probability is low.

jeanconn:
Reduce acq spoiler from red to orange and add new red warn for guide spoiler

There's a bit of code duplication here (and for a BOT star there could be warning duplication if the star has a spoiler within 25 arcsec) but this seems a relatively straightforward way to separate out the spoiler warnings.

The question of re-evaluating the usefulness of the spoiler warnings in the proseco era is a different issue.

Closes #236 

#246 already closed

### DIFF
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -1675,10 +1675,7 @@ sub check_star_catalog {
 	    }
 
             # Spoiler star in track box ACA-022
-            if (($type =~ /BOT|GUI/)
-                    and ($dz < $self->{dither_guide}->{ampl_p} + 25)
-                    and ($dy < $self->{dither_guide}->{ampl_y} + 25)
-                    and ($dm > -1.0)){
+            if (($type =~ /BOT|GUI/) and ($dz < 25) and ($dy < 25) and ($dm > -1.0)){
 		my $warn = sprintf("$alarm [%2d] Spoiler. %10d: " .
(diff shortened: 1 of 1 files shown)


jeanconn:
I thought maybe we'd get this in and cut starcheck 13.1 for a release this week/today along with https://github.com/sot/skare3/pull/120 .

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1674,13 +1674,24 @@ sub check_star_catalog {
 		else { push @yellow_warn, $warn }
 	    }
 
-	    # Star within search box + search error and within 1.0 mags 
-	    if ($type ne 'MON' and $dz < $halfw + $search_err and $dy < $halfw + $search_err and $dm > -1.0) {
+            # Spoiler star in track box ACA-022
+            if (($type =~ /BOT|GUI/)
+                    and ($dz < $self->{dither_guide}->{ampl_p} + 25)
```
taldcroft: I don't think that dither should be added in here since the two stars move in lock step.  Imagine a 64x64 dither and a spoiler at 40 arcsec away.  That is definitely fine.

jeanconn: Oh good point.  In a dither habit.

PR #304, opened 2019-03-26
WIP: Add option to override kadi dynamic state dither with state from file

jeanconn:
Add option to override kadi dynamic state dither with state from file

I've used this for a test run, today, on JUN2716T using this as the dither state file:
```
fido: more jun2716t_dither_state.json
{
    "dither": "DISA",
    "dither_ampl_pitch": 0,
    "dither_ampl_yaw": 0,
    "dither_period_pitch": 0,
    "dither_period_yaw": 0,
    "time": 582574205.235
}
```

For the disable dither state, it doesn't really matter what the dither parameters are (because we generally use an AODITPAR to set explicitly before enabling dither), but these 6 keys should have values to keep Parse_CM_File happy.

### DIFF
--- a/starcheck/src/starcheck.pl
+++ b/starcheck/src/starcheck.pl
@@ -49,6 +49,7 @@ use Inline Python => q{
 
 import os
 import traceback
+import json
 
 from Chandra.Time import DateTime
 from chandra_aca.star_probs import set_acq_model_ms_filter
@@ -153,6 +154,8 @@ def get_run_start_time(run_start_time, backstop_start):
     return ref_time.date
 
 
+def dither_from_json(file):
+    return json.load(open(file, 'r'))
 };
 
 
@@ -174,11 +177,12 @@ my %par = (dir  => '.',
                    agasc_file => "${SKA}/data/agasc/proseco_agasc_1p7.h5",
 		   config_file => "characteristics.yaml",
 		   fid_char => "fid_CHARACTERISTICS",
+                   dith_state_file => "",
 		   );
 
 
-GetOptions( \%par, 
-			'help', 
+GetOptions( \%par,
+			'help',
 			'dir=s',
 			'out=s',
 			'plot!',
@@ -190,7 +194,8 @@ GetOptions( \%par,
 			'sc_data=s',
 			'fid_char=s',
 			'config_file=s',
-                        'run_start_time=s',
+            'run_start_time=s',
+            'dith_state_file=s',
 			) ||
     exit( 1 );
 
@@ -213,7 +218,7 @@ usage( 1 )
 # Find backstop, guide star summary, OR, and maneuver files.
 my %input_files = ();
 
-# for split loads directory and prefix configuration 
+# for split loads directory and prefix configuration
 my $sosa_dir_slash = $par{vehicle} ? "vehicle/" : "";
 my $sosa_prefix = $par{vehicle} ? "V_" : "";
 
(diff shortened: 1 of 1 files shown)


taldcroft:
Looks generally good, but there should probably be some validation that the file has the right keys, and more importantly some docs so that when I need this in 3 years then I'll manage.  Maybe just copying the example file into the `-help` page and also printing an error message if validation fails with that example.

jeanconn:
It looks like for SEP0619B, we ran into a condition where the kadi commanded states dither state didn't match the History
```
Dither status in kadi commands does not match DITHER history
```
And the dither amplitudes were left undefined when passed to proseco to determine P2 values causing a TypeError
```TypeError: unsupported operand type(s) for +: 'NoneType' and 'int' at line 45
 at ./starcheck/src/starcheck.pl line 45.
	main::__ANON__("TypeError: unsupported operand type(s) for +: 'NoneType' and "...) called at (eval 80) line 3
	Ska::Starcheck::Obsid::proseco_probs("__main__", "proseco_probs", HASH(0x4888800)) called at starcheck/src/lib/Ska/Starcheck/Obsid.pm line 2997
	Ska::Starcheck::Obsid::set_proseco_probs_and_check_P2(Ska::Starcheck::Obsid=HASH(0x6de59d0)) called at ./starcheck/src/starcheck.pl line 690
```
Whichever way we decide to go, we should probably configure so either History or kadi commands "wins" and makes some kind of dither states for starcheck to proceed (with big red warnings or whatever).

taldcroft:
I vote for kadi winning, since that is the product that we control for a flight review.  History should become something that is a sanity check but not used in the starcheck review.

jeanconn:
I agree with kadi commanded states winning (or an override dither state from this PR).

taldcroft:
I rebased this.  I used VS code to help with the rebase and it works very well to better visualize the changes and in several cases accept both HEAD and incoming changes.  On the down side it automatically removes trailing white space, generating a lot of unrelated diffs.

taldcroft:
Note in the GitHub diff there is a setting button to remove whitespace diffs. With that, this rebase appears the same as #345.

taldcroft:
Though after all that I would still rather fix this problem correctly instead of putting in a back-door for handling a situation that the flight code (run by SOT MP) should handle.

jeanconn:
I think we should still have an override for weird interrupts.  There are other things that aren't ideal... I assume that the kadi cmds cron job will pick up my update to dither cmds at some point... but what point would that be?  Sometimes we're in a little bit of a hurry.

jeanconn:
And sure, I did not think that rebasing would be a challenge, just looking at the piece of code and having minutes it was cleaner for me to just copy the edits I wanted to make a new branch to represent exactly what I ran for MAY2620B.  I did not really care which one got used and figured I'd just close one or the other.  If you prefer the rebased one that is fine.

jeanconn:
And of course still want tools to do nonload commands for us.  I suppose I need to stop the previous CTI at some point etc...

jeanconn:
PS In our previous discussion, it was "let kadi win".  Do we still want that?  If so, I'm fine with that, though thus far History has been more right than kadi continuity due to my manual edits.  If we set the kadi status to "win",  I'd just want to put a big red warning at the top of starcheck outputs indicating a mismatch with History if that happened (instead of starcheck just dying trying to calculate p_acq on None dither amplitudes).

jeanconn:
This is overcome by events.  With https://github.com/sot/starcheck/pull/390 the kadi state is authoritative and with command states v2 one could use a KADI_SCENARIO for the rare case of needing to set dither to something different for the starcheck.

PR #309, opened 2019-06-07
Use get_effective_t_ccd

jeanconn:
Use get_effective_t_ccd

Starcheck needs the effective_t_ccd a bunch of places anyway, so I'm not sure if we win anything by doing the get_aca_catalog transition.  I'll work that in another branch.

Of course, if I'm calling get_effective_t_ccd, the trick is to confirm I'm not accidentally calling it twice on any input temperature.

### DIFF
--- a/starcheck/data/characteristics.yaml
+++ b/starcheck/data/characteristics.yaml
@@ -62,8 +62,8 @@ no_starcat_oflsid:
    - RDE
    - DC_T
 
-ccd_temp_yellow_limit: -10.2
-ccd_temp_red_limit: -9.5
+ccd_temp_yellow_limit: -9.5
+ccd_temp_red_limit: -8.5
 
-n100_warm_frac_default: .15
+n100_warm_frac_default: .20
 
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -33,6 +33,7 @@ from Ska.quatutil import radec2yagzag
 import agasc
 
 from proseco.core import ACABox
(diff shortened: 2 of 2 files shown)


jeanconn:
Though all the uglier parts are probably where we need review...

taldcroft:
Just wondering if it is possible to use the `aca` object from `get_aca_catalog` to populate the Obsid object with guide and acq effective temperatures, and then use those where appropriate.  I.e. don't change the low-level helper functions, but have the routines that call them decide which temperature to pass.  This should remove any possibility of double-effectiving because it is never directly called in starcheck.  Then in the info statement just check if t_ccd != t_ccd_eff and issue a statement if so.  That decouples the details of computing effective temperature from the perl code to issue an info statement.

Anyway, having said that, what you have here will work just fine and I'm OK with it.  PITEOG.

taldcroft:
Oh, and if you go with the current approach then I agree you should just close #310.

jeanconn:
Though #310 just moves things in the right direction and could go as well.  It might be a good first step toward working with an ACACatalogTable in starcheck anyway.

jeanconn:
Also, whatever we do what do you think about testing?   I have played both with changing all of the temperatures (just adding 2 degrees in the original assignments) and played with changing the planning limits to something lower to use old schedules.

taldcroft:
That testing is fine.

jeanconn:
I looked more at getting the effective temperatures back from the ACACatalogTable, but given we would like this for this week, I'm thinking of just not fiddling more with changing the API of anything that goes across the Perl / Python divide.  Is this approved for now with the PITEOG in mind?

PR #310, opened 2019-06-07
Replace get_acq_catalog with get_aca_catalog and cleanup

jeanconn:
Replace get_acq_catalog with get_aca_catalog and cleanup


jeanconn:
This is a PR against #309

PR #318, opened 2019-06-12
Make the sandbox launcher work on standalone Ska3

taldcroft:
This seems like a safe change in that it doesn't impact production starcheck.

### DIFF
--- a/sandbox_starcheck
+++ b/sandbox_starcheck
@@ -1,7 +1,9 @@
 #!/bin/bash
 
 DIR=`dirname $0`
-export PYTHONHOME=$SKA
+PYTHON_EXE=`which python`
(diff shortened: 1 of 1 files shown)


sandbox_starcheck
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
 
 DIR=`dirname $0`
-export PYTHONHOME=$SKA
+PYTHON_EXE=`which python`
```
jeanconn: I think there are some options for `which` in bash (maybe to skip aliases and such) but we probably don't need to care.

taldcroft: The don't-care seems about right to me.

PR #321, opened 2019-06-21
Update checklist and starcheck for 11.2 maxmag limit

### LINKED ISSUE #320: Update for new maxmag 11.2 clipping

jeanconn:
Update checklist and starcheck for 11.2 maxmag limit

Closes #320

### DIFF
--- a/starcheck/src/aca_load_review_cl.rst
+++ b/starcheck/src/aca_load_review_cl.rst
@@ -330,7 +330,7 @@ Checks
 +-------+------------------+-+-+-------------------------------------------+----+----------------+
 |       |                  |X|X|                                           |    |AS: Possible    |
 |       |                  | | |                                           |    |Bright Star     |
-|ACA-041|Magnitude         | | |Slot MAXMAG (faint limit) - star MAG >= 1.4|n/a |Hold, GS:       |
+|ACA-041|Magnitude         | | |Slot MAXMAG (faint limit) - star MAG >= 0.3|n/a |Hold, GS:       |
 |       |                  | | |                                           |    |Reduced aspect  |
 |       |                  | | |                                           |    |quality         |
 +-------+------------------+-+-+-------------------------------------------+----+----------------+
--- a/starcheck/src/aca_load_review_cl.rst
+++ b/starcheck/src/aca_load_review_cl.rst
@@ -330,7 +330,7 @@ Checks
 +-------+------------------+-+-+-------------------------------------------+----+----------------+
 |       |                  |X|X|                                           |    |AS: Reduced     |
 |       |                  | | |                                           |    |acq probability |
-|       |                  | | |                                           |    |on the star     |
+|       |                  | | |                                           |    |of star     |
 |ACA-041|Magnitude         | | |Slot MAXMAG (faint limit) - star MAG >= 0.3|n/a |GS: Increased   |
 |       |                  | | |                                           |    |risk of loss of |
 |       |                  | | |                                           |    |track of star   |
--- a/starcheck/src/aca_load_review_cl.rst
+++ b/starcheck/src/aca_load_review_cl.rst
@@ -328,11 +328,12 @@ Checks
 |ACA-040|Monitor commanding|X|X|Dither is disabled and enabled with correct|n/a |Failure to track|
 |       |                  | | |timing                                     |    |                |
 +-------+------------------+-+-+-------------------------------------------+----+----------------+
-|       |                  |X|X|                                           |    |AS: Possible    |
-|       |                  | | |                                           |    |Bright Star     |
-|ACA-041|Magnitude         | | |Slot MAXMAG (faint limit) - star MAG >= 0.3|n/a |Hold, GS:       |
-|       |                  | | |                                           |    |Reduced aspect  |
-|       |                  | | |                                           |    |quality         |
+|       |                  |X|X|                                           |    |AS: Reduced     |
+|       |                  | | |                                           |    |acq probability |
+|       |                  | | |                                           |    |on the star     |
+|ACA-041|Magnitude         | | |Slot MAXMAG (faint limit) - star MAG >= 0.3|n/a |GS: Increased   |
+|       |                  | | |                                           |    |risk of loss of |
+|       |                  | | |                                           |    |track of star   |
 +-------+------------------+-+-+-------------------------------------------+----+----------------+
 |ACA-042|AGASC requirements|X|X|An AGASC star exists within ID_DIST_LIMIT  |n/a |Failure to track|
 |       |                  | | |(1.5as) of the center of each search box   |    |star            |
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -1593,17 +1593,18 @@ sub check_star_catalog {
                 if (($maxmag - $mag) < 0.3){
                     push @warn, sprintf "[%2d] Magnitude.  MAXMAG - MAG < 0.3\n", $i;
                 }
-                # And this is a sanity check on the current maxmag clipping at 11.2.
+                # This is a check that maxmag for each slot is as-expected.
+                # Note that for stars with large mag err (like color 1.5 stars) this will throw
+                # a warning.
                 my $rounded_maxmag = sprintf("%.2f", $maxmag);
                 my $expected_maxmag = min($mag + 1.5, 11.2);
 		if (abs($expected_maxmag - $rounded_maxmag) > 0.1){
-                    push @orange_warn,
+                    push @yellow_warn,
                         sprintf "[%2d] Magnitude.  MAXMAG %.2f not within 0.1 mag of %.2f \n",
                     $i, $rounded_maxmag, $expected_maxmag;
 		}
 	    }
 	}
-	
 
 	# Search box too large ACA-018
 	if ($type ne 'MON' and $c->{"HALFW$i"} > 200) {
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -1214,8 +1214,6 @@ sub check_star_catalog {
     my $col_sep_dist = 50;	# Common column pixel separation
     my $col_sep_mag  = 4.5;	# Common column mag separation (from ODB_MIN_COL_MAG_G)
 
-    my $mag_faint_slot_diff = 1.4; # used in slot test like:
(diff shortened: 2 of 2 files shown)


starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1593,13 +1590,31 @@ sub check_star_catalog {
...
+		    push @warn, sprintf "$alarm [%2d] Magnitude.  MAXMAG - MAG < 1.4 \n",$i;
+                }
+                elsif (($type =~ /BOT|GUI/) and ($maxmag - $mag < $max_mag_clip - $max_guide_mag)){
+		    push @orange_warn, sprintf "$alarm [%2d] Magnitude.  Guide star MAXMAG - MAG < %.1f \n",$i,
+                        $max_mag_clip - $max_guide_mag;
+                }
+                elsif (($type =~ /ACQ/) and ($maxmag - $mag < $max_mag_clip - $max_acq_mag)){
+		    push @orange_warn, sprintf "$alarm [%2d] Magnitude.  Acquisition star MAXMAG - MAG < %.1f \n",$i,
+                        $max_mag_clip - $max_acq_mag;
+                }
```
jeanconn: Honestly, we could go with either the first check OR the latter two as sanity checks.  None of this seems really necessary, but is also not going to show up in the output.

starcheck/src/aca_load_review_cl.rst
```diff
@@ -330,7 +330,7 @@ Checks
 +-------+------------------+-+-+-------------------------------------------+----+----------------+
 |       |                  |X|X|                                           |    |AS: Possible    |
 |       |                  | | |                                           |    |Bright Star     |
-|ACA-041|Magnitude         | | |Slot MAXMAG (faint limit) - star MAG >= 1.4|n/a |Hold, GS:       |
+|ACA-041|Magnitude         | | |Slot MAXMAG (faint limit) >= star MAG      |n/a |Hold, GS:       |
```
jeanconn: Thoughts?

jeanconn: OK.  Per our discussion I'm going to chuck this and go  MAXMAG is more than MAG + 0.5 or MAXMAG == 11.2 .  

MAXMAG >= min(MAG + 0.5, 11.2)

jeanconn: I suppose in our current operational posture it should also be 11.2 >= MAXMAG >= min(MAG + 0.5, 11.2), but I don't think we care about adding a check that MAXMAG is not greater than 11.2

taldcroft: What about just changing 1.4 to 0.3.  So basically the checklist item should never change again and it is a simple, hard limit.  Starcheck is free to be more subtle.  Then add a second condition MAXMAG <= 11.2.  That is another hard limit that we are very unlikely to change.

jeanconn:
Did you mean 0.2 hard limit?

starcheck/src/aca_load_review_cl.rst
jeanconn: Did you mean 0.2 hard limit? (putting this comment back in the thread).  I had read 11.0 as the non-temperature scaled ACQ star faint limit from here:

https://github.com/sot/proseco/blob/ad6f861d0fd9eb27cc7034731ed758c9b875f84b/proseco/acq.py#L285

And by the time we got to maxmag - mag >= 0.2, I had just proposed changing the checklist item to just having no margin (maxmag >= mag) which is how this PR started.  But then I was sanity checking in starcheck to see that we were still getting reasonable values.

taldcroft: With the 11.2 maxmag we could consider pushing down the acq selection limit.  The reason that limit is not temperature scaled is that there is an accurate temperature-dependent model of star acq performance, so the upper bar can just be set at an arbitrary high but not ridiculously high limit.    We could drop it down to 10.9 without any impact, except for wasting time re-validating.  Or we could just put a 0.3 value in the checklist and never notice, in practice.  I think that we would not ever want less than 0.3 mag margin.

jeanconn: I'm fine with updating the checklist to use 0.3 mag margin.  If we go with that I think it would also make sense to update the "risks" section, to indicate that less than that could reduce acq probability or ability to track that single star.  ACA-041 looks scarier than it needs to be with "bright star hold" and "reduced aspect quality".

starcheck/src/aca_load_review_cl.rst
```diff
@@ -328,11 +328,12 @@ Checks
 |ACA-040|Monitor commanding|X|X|Dither is disabled and enabled with correct|n/a |Failure to track|
 |       |                  | | |timing                                     |    |                |
 +-------+------------------+-+-+-------------------------------------------+----+----------------+
-|       |                  |X|X|                                           |    |AS: Possible    |
-|       |                  | | |                                           |    |Bright Star     |
-|ACA-041|Magnitude         | | |Slot MAXMAG (faint limit) - star MAG >= 0.3|n/a |Hold, GS:       |
-|       |                  | | |                                           |    |Reduced aspect  |
-|       |                  | | |                                           |    |quality         |
+|       |                  |X|X|                                           |    |AS: Reduced     |
+|       |                  | | |                                           |    |acq probability |
```
jeanconn: Regarding the implications column, I think technically this would suggest that we should have starcheck set P_ACQ to 0 for acq stars with MAXMAG - MAG < 0.3 .  But I can PITEOG on this.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: Since I'll need to figure out rounding in Perl again, why 2 digits?  I would figure 1 decimal.

taldcroft: I just figured that if there really was a maxmag of 11.249 we would want to know.  I'm not sure of the precision of maxmag in backstop, but thought that two decimal places is safe and 11.2049 would more likely just be numerics.  Maybe 3 decimals is also fine.

I thought round is perl is done with sprintf and that horrible string-can-be-a-number?

jeanconn: In backstop, per my test report, we have 1.12031250e+01 .

And I thought sprintf did truncation but not rounding?

taldcroft: My understanding is that the C-based sprintf format codes like 5.2f are rounding.

So it looks like it needs to be 1-digit in any case.  I guess that value doesn't have many bits, like a 16-bit float?  Anyway, no matter.

taldcroft: Perl rounds according to SO:
https://stackoverflow.com/questions/9781774/truncate-not-round-decimal-places-in-sprintf

jeanconn: I got confused about the 2 digits / 1 decimal started with 2.  Am I getting warmer?

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1586,15 +1584,22 @@ sub check_star_catalog {
...
-		if (($maxmag - $mag) < $mag_faint_slot_diff){
-		    my $slot_diff = $maxmag - $mag;
-		    push @warn, sprintf "[%2d] Magnitude.  MAXMAG - MAG = %1.2f < $mag_faint_slot_diff \n",$i,$slot_diff;
+                # This is an explicit check of ACA-041
+                if (($maxmag - $mag) < 0.3){
+                    push @warn, sprintf "[%2d] Magnitude.  MAXMAG - MAG < 0.3\n", $i;
+                }
+                # And this is a sanity check on the current maxmag clipping at 11.2.
+                my $rounded_maxmag = sprintf("%.2f", $maxmag);
+                my $expected_maxmag = min($mag + 1.5, 11.2);
```
jeanconn: It looks like this is throwing false positives (I think out of the matlab tools for larger mag errors, the original maxmag value would have been more than $mag + 1.5).

taldcroft: That's an interesting question if this should be considered a False positive.  I.e. do we actually want to allow `maxmag > mag + 1.5` even for stars with large mag err?  My first reaction is that we should just stick with 1.5 no matter what.  Of course MATLAB is doing what it does now, so this warning serves as a notice of a large maxmag.  Remember that strictly speaking, starcheck warnings are evaluated by the ACA reviewer and a catalog can be passed at ACA reviewer discretion if it does not violate any FDB-controlled checklist items.

jeanconn: Regarding "false positive", I just didn't think we wanted to add a new check that would add possibly a bunch of orange warning that we aren't going to care about.  For example,  For example star 1039018944 in 48042 looks to be a color=1.5 star that is being handled with a larger mag error on the matlab side and ends up with a 11.20 clipped value.  I thought the "green" handling of it as a 1.5 star was probably enough:

ttps://icxc.cfa.harvard.edu/mp/mplogs/2019/JUL0719/oflsa/starcheck.html#obsid48042

and it does not need 

>> WARNING : [ 6] Magnitude.  MAXMAG 11.20 not within 0.1 mag of 10.28

taldcroft: Fair enough.  My point is that MATLAB should not be giving that star a maxmag that is so large.  Previously that was just fine and we didn't care, but now that could actually negatively impact acquisition probability (compared to the alternative of maxmag = mag + 1.5).  

Some visibility into this might be OK as long as it is not happening *too often*.  (No, I don't have a strict definition of what that means).  But I also agree (and have a long history of advocating) that non-actionable warnings should be avoided.  Can you make it an info?  I think we should be working toward changing MATLAB handling of maxmag, but that may take some consideration of cases like this.

jeanconn: I agree on prioritizing figuring out what to actually do with the handling of maxmag .  And with regard to the warning coming up, it looks like it will just do that a lot more when we have more color=1.5 stars (which are often whole fields).  With regard to this new warning, sure I can go with blue / INFO or green warning / CAUTION.

jeanconn: OK, I've set to green / CAUTION (consistent with the color of the row for the color=1.5 stars anyway).

jeanconn:
I've updated the text *one* more time so the warning now looks like
```
>> CAUTION : [ 6] Magnitude.  MAXMAG 11.20 not within 0.1 mag of 10.28. (MAXMAG-MAG=2.42) 
```

PR #323, opened 2019-06-25
Update warning text to include CAUTION and CRITICAL

### LINKED ISSUE #315: Print warnings within a severity in some kind of order
Right now I'm seeing extra diffs because warnings are printed in no order within a "severity".

jeanconn:
Update warning text to include CAUTION and CRITICAL .  

This also sorts the warnings and 

Closes #315 

This will need an updated starcheck_parser in mica.

### DIFF
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -2281,29 +2281,29 @@ sub print_report {
 
     if (@{$self->{warn}}) {
 	$o .= "${red_font_start}";
-	foreach (@{$self->{warn}}) {
-	    $o .= ">> CRITICAL:" . $_;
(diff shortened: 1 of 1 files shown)


taldcroft:
> This will need an updated starcheck_parser in mica.

About the updated starcheck parser, I've forgotten again why we are not just using the JSON file.

jeanconn:
Because of limited resources?  The starcheck_parser change is a few lines. Actually using the JSON file is a significant undertaking both to really put the things in it that we need (it is mostly just a mishmash now), put them in there in a useful structure, and then adapt all of the database code in mica.starcheck to work with that new structure.  I'd like to do all that (!); just have not seen it as a priority.  I've gotten as far as taking the action of "I should really look at that at some point".

taldcroft:
About the JSON, thanks for refreshing my memory.  Your explanation makes sense.

PR #324, opened 2019-06-27
Remove regression test files from starcheck source repo

### LINKED ISSUE #316: Finish pulling out regression outputs
I got started pulling out the regression outputs in https://github.com/sot/starcheck_regress , but never got around to finishing.  I think this just needs some setup in the starcheck Makefile and some notes/docs for process.

I think that should be re-baselined at starcheck 13 (or now 13.2) and all of the old regression outputs should be tossed.

taldcroft:
Can we just do this?  I see #316 as a WIP.  It seems like we can just make  https://github.com/sot/starcheck_regress start from today as a *copy* of the current release regression files. In the unlikely event that the regression outputs for old releases are actually needed then they are still in the starcheck repo.  PITEOG.

Closes #316


jeanconn:
We can just do this, but won't they still show up in github search?  I thought that was your second big issue with having them in the repo.  I figured removing them from history would be fine once I was sure about a baseline.

taldcroft:
I think that github search just indexes master.  It would be a mess if searching showed every instance of a string if it occurred in any past commit.

taldcroft:
BTW, if you wanted all regress history in one place, then an option would be to clone `starcheck` to `starcheck_regress`.  Then just delete all the source files there in one commit.  They would be hanging out still in the commit history, but that wouldn't really bother anything.

jeanconn:
OK.  I'm fine with merging this and if it doesn't fix the github search will strip from history.

taldcroft:
:tada:  I think this had the intended effect!  I can now search on "spoiler" and get only relevant results.

PR #325, opened 2019-06-27
Remove double-printed labels on a couple of obsid warnings

jeanconn:
Remove double-printed labels on a couple of obsid warnings


jeanconn:
I missed these in #323

PR #326, opened 2019-06-27
Move starcheck parser into starcheck

taldcroft:
It would seem to make more sense here than in mica.  This requires https://github.com/sot/mica/pull/201.

### DIFF
--- a/starcheck/parser.py
+++ b/starcheck/parser.py
@@ -1,0 +1,328 @@
+#!/usr/bin/env
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+
+import re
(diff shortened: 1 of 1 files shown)


taldcroft:
Tested this by running annie and proseco tests, which use `read_starcheck`.  Pass.

jeanconn:
> It would seem to make more sense here than in mica. 

Hah.  It did start here (the Perl version is still hanging around in here) but then I kept wanting to change the parsing to work with the database ingest and didn't want to keep cutting starcheck releases to match.  Will see how it goes.  Would like to re-examine our use cases at some point too.

PR #319, opened 2019-06-14
Use kadi dynamic commanded states instead of cmd_states

taldcroft:
This runs without error on JUN1019A and JUN1719A and appears to have the desired effect of properly accounting for eclipse times.  It needs real testing, of course.

It is based on #318 for convenience and requires https://github.com/sot/kadi/pull/131.

I took the liberty of fixing some PEP8 stuff and simplifying a few things here and there.

PR #329, opened 2019-07-12
Add a run_start_time and change initial state determination per #319

jeanconn:
Add a run_start_time for regression and change the determination of
the reference / initial / seed state with temperature.  The previous
code used a state at least 3 days before backstop tstart and at least
one our before the last available telemetry.  The new code uses a state
within 1400 seconds of the end of available telemetry (where the
available telemetry is limited by either real availability, backstop
tstart, or the new 'run_start_time' option).

### DIFF
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -101,6 +101,7 @@ def get_options():
 def get_ccd_temps(oflsdir, outdir='out',
                   json_obsids=sys.stdin,
                   model_spec=None, char_file=None, orlist=None,
+                  run_start_time=None,
                   verbose=1, **kwargs):
     """
     Using the cmds and cmd_states sybase databases, available telemetry, and
@@ -118,16 +119,19 @@ def get_ccd_temps(oflsdir, outdir='out',
     if not os.path.exists(outdir):
         os.mkdir(outdir)
 
+    run_start_time = DateTime(run_start_time)
     config_logging(outdir, verbose)
 
     # Store info relevant to processing for use in outputs
     proc = {'run_user': os.environ['USER'],
-            'run_time': time.ctime(),
+            'actual_run_time': time.ctime(),
+            'run_start_time': run_start_time,
             'errors': []}
     logger.info('##############################'
                 '#######################################')
     logger.info('# %s run at %s by %s'
-                % (TASK_NAME, proc['run_time'], proc['run_user']))
+                % (TASK_NAME, proc['actual_run_time'], proc['run_user']))
+    logger.info("# Using {} as run_start_time".format(run_start_time.date))
     logger.info('# {} version = {}'.format(TASK_NAME, VERSION))
     logger.info('###############################'
                 '######################################\n')
@@ -153,8 +157,11 @@ def get_ccd_temps(oflsdir, outdir='out',
     proc['datestart'] = DateTime(tstart).date
     proc['datestop'] = DateTime(tstop).date
 
-    # Get temperature telemetry for 30 days prior to min(tstart, NOW)
-    tlm = get_telem_values(min(tstart, tnow),
+
+    # Get temperature telemetry for 30 days prior to
+    # min(tstart, NOW, run_start_time) where run_start_time is basically
+    # a mock NOW for regression testing.
+    tlm = get_telem_values(min(tstart, tnow, run_start_time.secs),
                            ['aacccdpt', 'pitch'],
                            days=30)
 
@@ -292,22 +299,22 @@ def get_week_states(tstart, tstop, bs_cmds, tlm):
     :param tlm: available pitch and aacccdpt telemetry recarray from fetch
     :returns: numpy recarray of states
     """
-    cstates = Table(get_cmd_states.fetch_states(DateTime(tstart) - 30,
+
+    # Get temperature data at the end of available telemetry
+    ok = tlm['date'] > tlm['date'][-1] - 1400
+    init_aacccdpt = np.mean(tlm['aacccdpt'][ok])
+    init_tlm_time = np.mean(tlm['date'][ok])
+
+    # Get states at and after that time
+    cstates = Table(get_cmd_states.fetch_states(init_tlm_time,
                                                 tstop,
                                                 vals=['obsid',
                                                       'pitch',
                                                       'q1', 'q2', 'q3', 'q4']))
-    # Get the last state at least 3 days before tstart and at least one hour
-    # before the last available telemetry
-    cstate0 = cstates[(cstates['tstart'] < (DateTime(tstart) - 3).secs)
-                      & (cstates['tstart'] < (tlm[-1]['date'] - 3600))][-1]
-    # get temperature data in a range around that initial state
-    ok = ((tlm['date'] >= cstate0['tstart'] - 700) &
-          (tlm['date'] <= cstate0['tstart'] + 700))
-    init_aacccdpt = np.mean(tlm['aacccdpt'][ok])
+    cstate0 = cstates[0]
+    pre_bs_states = cstates[(cstates['tstart'] >= cstate0['tstart']) &
+                            (cstates['tstart'] < tstart)]
 
-    pre_bs_states = cstates[(cstates['tstart'] >= cstate0['tstart'])
-                            & (cstates['tstart'] < tstart)]
     # cmd_states.get_states needs an initial state dictionary, so
     # construct one from the last pre-backstop state
     last_pre_bs_state = {col: pre_bs_states[-1][col]
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -113,6 +113,10 @@ def get_ccd_temps(oflsdir, outdir='out',
     :param json_obsids: file-like object or string containing JSON of
                         starcheck Obsid objects
     :param model_spec: xija ACA model specification
+    :param run_start_time: Chandra.Time date used as a reference time to determine initial
+                     seed state with temperature telemetry.  The initial seed state will
+                     be at the end of available telemetry that is also before run_start_time,
+                     before the beginning of backstop cmds, and before "now".
     :param verbose: Verbosity (0=quiet, 1=normal, 2=debug)
     :returns: JSON dictionary of labeled dwell intervals with max temperatures
     """
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -124,14 +124,14 @@ def get_ccd_temps(oflsdir, outdir='out',
 
     # Store info relevant to processing for use in outputs
     proc = {'run_user': os.environ['USER'],
-            'actual_run_time': time.ctime(),
+            'execution_time': time.ctime(),
             'run_start_time': run_start_time,
             'errors': []}
     logger.info('##############################'
                 '#######################################')
     logger.info('# %s run at %s by %s'
-                % (TASK_NAME, proc['actual_run_time'], proc['run_user']))
-    logger.info("# Using {} as run_start_time".format(run_start_time.date))
+                % (TASK_NAME, proc['execution_time'], proc['run_user']))
+    logger.info("# Continuity run_start_time = {}".format(run_start_time.date))
     logger.info('# {} version = {}'.format(TASK_NAME, VERSION))
     logger.info('###############################'
                 '######################################\n')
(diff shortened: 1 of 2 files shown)


starcheck/calc_ccd_temps.py
```diff
@@ -292,22 +299,22 @@ def get_week_states(tstart, tstop, bs_cmds, tlm):
...
+
+    # Get states at and after that time
+    cstates = Table(get_cmd_states.fetch_states(init_tlm_time,
                                                 tstop,
                                                 vals=['obsid',
                                                       'pitch',
                                                       'q1', 'q2', 'q3', 'q4']))
-    # Get the last state at least 3 days before tstart and at least one hour
-    # before the last available telemetry
-    cstate0 = cstates[(cstates['tstart'] < (DateTime(tstart) - 3).secs)
```
jeanconn: I do have more temperature diffs in the regression set than I thought I would have, and I suspect most of that is due to removing this "3 days before backstop tstart" rule when running these regression weeks.  Do we need to quantify this (show the diffs are mostly at the beginning of schedules) or just run a couple of non-eclipse weeks and show they match MCC well enough?

taldcroft: Run a couple of non-eclipse loads and show diffs less than 0.25 C (as you mentioned in a different thread).  There are still some state diffs from ORviewer vs. kadi or cmd_states.

jeanconn: Yes.  I looked at mar1119 and mar2519 and they look good to me.

mar1119b_diff
mar2519a_diff

jeanconn: I don't entirely understand even the small diffs, but for the ACA thermal model I was comfortable with these as at the level of "don't care".

starcheck/src/starcheck.pl
```diff
@@ -108,6 +111,47 @@
...
+    :returns: YYYY:DOY string of reference run start time
+    """
+
+    run_start = de_bytestr(run_start)
+    ref_time = DateTime(run_start)
+    # For the special case where run_start casts as a float
+    # check to see if it is negative and if so, override the reference
+    # time with a time run_start days back from backstop start
+    try:
+        run_start = float(run_start)
```
jeanconn: @taldcroft Is this about the level of "hacky" that you had in mind?

starcheck/src/starcheck.pl
```diff
@@ -48,6 +48,7 @@
 use Inline Python => q{
 
 import os
+import traceback
```
taldcroft: Unused import?

starcheck/src/starcheck.pl
taldcroft: Looks fine.  Just some nits:
- The first `ref_time = DateTime(run_start)` should be in the `except` block since it should only run if the input is not a float.
- Do `except TypeError:`
- Put the `if run_start < 0:` stuff into an `else:` block.  This part should not be excepted from errors, you always want to limit exception catching to just the code in question.
- There should probably be an `else:` block on the `if run_start < 0` which raises a ValueError.  The current behavior for positive values will give an unexpected result (i.e. a date near 1998:001) and likely make a difficult-to-understand crash later.

starcheck/src/starcheck.pl
```diff
@@ -108,6 +109,45 @@
...
+
+    run_start = de_bytestr(run_start)
+    ref_time = DateTime(run_start)
+    # For the special case where run_start casts as a float
+    # check to see if it is negative and if so, override the reference
+    # time with a time run_start days back from backstop start
+    try:
+        run_start = float(run_start)
+        if run_start < 0:
+            backstop_datetime = DateTime(de_bytestr(backstop_start))
```
taldcroft: The `de_bytestr` processing should always be a separate line at the top of the function.  Separate the function input wrangling from the actual processing.  Then this just becomes one line:
```
ref_time = DateTime(backstop_start) + run_start  # Already a DateTime, no recast required
```

starcheck/calc_ccd_temps.py
```diff
@@ -118,16 +119,19 @@ def get_ccd_temps(oflsdir, outdir='out',
     if not os.path.exists(outdir):
         os.mkdir(outdir)
 
+    run_start_time = DateTime(run_start_time)
     config_logging(outdir, verbose)
 
     # Store info relevant to processing for use in outputs
     proc = {'run_user': os.environ['USER'],
-            'run_time': time.ctime(),
+            'actual_run_time': time.ctime(),
```
taldcroft: If you are changing the name, how about `execution_time`?

starcheck/calc_ccd_temps.py
```diff
@@ -118,16 +119,19 @@ def get_ccd_temps(oflsdir, outdir='out',
...
+            'run_start_time': run_start_time,
             'errors': []}
     logger.info('##############################'
                 '#######################################')
     logger.info('# %s run at %s by %s'
-                % (TASK_NAME, proc['run_time'], proc['run_user']))
+                % (TASK_NAME, proc['actual_run_time'], proc['run_user']))
+    logger.info("# Using {} as run_start_time".format(run_start_time.date))
```
taldcroft: Maybe `# Continuity run_start_time = ...`?

starcheck/src/starcheck.pl
```diff
@@ -108,6 +109,45 @@
...
 
 
+def get_now_time():
+    """
+    Return the current time in YYYY:DOY format
+    """
+    return DateTime().date
+
+
+def run_start_time(run_start, backstop_start):
```
taldcroft: Call this `get_run_start_time` for consistency with the previous, and to avoid overloading `run_start_time` with two things, the variable name (in perl) and the function name in Python.  Then you can use `run_start_time` as the arg name here and avoid name changes.

starcheck/src/starcheck.pl
```diff
@@ -108,6 +109,45 @@
...
+    :param run_start: supplied run start time in a Chandra.Time format,
+                      negative numbers interpreted as "days back" relative
+                      to first backstop command
+    :param backstop_start: time of first backstop command
+    :returns: YYYY:DOY string of reference run start time
+    """
+
+    run_start = de_bytestr(run_start)
+    ref_time = DateTime(run_start)
+    # For the special case where run_start casts as a float
```
taldcroft: Always a blank line before a multiline comment.

starcheck/src/starcheck.pl
```diff
@@ -587,6 +628,11 @@ sub json_obsids{
 }
 
 
+# Set the thermal model run start time to be either the supplied
+# run_start_time or now.
+my $run_start_time = run_start_time($par{run_start_time}, $bs[0]->{date}) || get_now_time();
+
```
taldcroft: Only one blank line.

starcheck/src/starcheck.pl
```diff
@@ -587,6 +628,11 @@ sub json_obsids{
 }
 
 
+# Set the thermal model run start time to be either the supplied
+# run_start_time or now.
+my $run_start_time = run_start_time($par{run_start_time}, $bs[0]->{date}) || get_now_time();
```
taldcroft: The `get_run_start_time` function should do the entire job:
```
my $run_start_time = get_run_start_time($par{run_start_time}, $bs[0]->{date});
```
Then you don't really need the new `get_now_time()` function.

starcheck/calc_ccd_temps.py
```diff
@@ -292,22 +299,22 @@ def get_week_states(tstart, tstop, bs_cmds, tlm):
     :param tlm: available pitch and aacccdpt telemetry recarray from fetch
     :returns: numpy recarray of states
     """
-    cstates = Table(get_cmd_states.fetch_states(DateTime(tstart) - 30,
+
+    # Get temperature data at the end of available telemetry
```
taldcroft: I don't quite get all these diffs, which should be untouched by just adding `run_start_time`.  Is this about getting the regression tests looking OK?  Isn't this going to make rebasing `run_start_time_kadi` a problem?  Or I'm just confused about the plan with the two branches.

jeanconn: I don't understand this comment. 

Do you mean you don't want me to also change the definition of the initial state with temperature telemetry in this PR?  Because that's what I was asking about in our sit down and you were onboard with it then.  If not clear, I was trying to say, I originally wanted three PRS

1) with run_start_time
2) with run_start_time and change in definition of initial state with temperature
3) with run_start_time + change in definition + kadi

But once I settled on skipping to 2) with the understanding that you were OK without looking into great detail into the diffs caused by changing the definition of the initial state... as that was just your code and you thought it was the right thing.  That at least made it easy to diff the cmd_states and kadi_states version.

So these changes in this section are your changes to the definition of the initial state.

And they aren't completely about "getting the regression tests looking OK" because I don't know which "OK" you mean.  It shouldn't make rebasing `run_start_time_kadi` a problem, because the block is above the area of where the kadi change happens (but did make trying to rebase off your branch too much work).

jeanconn: I also tried to explain what was happening in this PR in the PR description.

starcheck/src/starcheck.pl
jeanconn: This one will get used in the kadi PR (which also wraps get_ccd_temps) so I didn't care about cutting it.

starcheck/src/starcheck.pl
jeanconn: Regarding: `ref_time = DateTime(run_start)` should be in the `except` block, I wasn't actually sure about this.  I was OK with a positive float being handled that way with the idea that the users gets what they get in that case.  What would you like them to get for that case?

jeanconn: Also, when getting to coding this, I thought it might make just as much sense to just have a separate option like `-days_back` instead of overloading `-run_start_time` this way, but...

jeanconn: OH, and I missed that you want a ValueError on a positive float.  My issue with that was I figured there was a whole area of "reasonableness" checking that we are not doing on these run_start_times, but if you want that specific case to get a ValueError that is fine.

taldcroft: Most likely user doesn't get what they want with a positive float.  This is what I meant about punting the error downstream in this case.
```
In [5]: DateTime('3').date
Out[5]: '1997:365:23:58:59.816'

In [6]: DateTime(3).date
Out[6]: '1997:365:23:58:59.816'
```

taldcroft: Since this is a testing thing, I don't care that much about raising `ValueError` in which case you just remove the `if run_start < 0` test.  But as shown above, user will end up querying Ska for 20 years of data if they put in a positive value.

jeanconn: And yes, I saw 3 would give you 1997.  I was somewhat wondering if you wanted to hack this even more so 3 or -3 would give you the same answer.

starcheck/calc_ccd_temps.py
taldcroft: OK, that makes sense.  I admit to not carefully reading the top description and I thought this was only what the title said (I've changed it now).

starcheck/src/starcheck.pl
taldcroft: No, because +3 is basically never a useful thing to do.  It will almost certainly break other parts of the code (you can't start from the middle of loads).  That's why ValueError is the correct response.

jeanconn: I meant same effective answer (-3 or 3 taking you 3 days before backstop start).

starcheck/calc_ccd_temps.py
jeanconn: OK.

starcheck/src/starcheck.pl
```diff
@@ -108,6 +109,44 @@
...
+
+    run_start_time = de_bytestr(run_start_time)
+    backstop_start = de_bytestr(backstop_start)
+
+    # For the special case where run_start_time casts as a float
+    # check to see if it is negative and if so, set the reference
+    # time to be a time run_start_time days back from backstop start
+    try:
+        run_start_time = float(run_start_time)
+    except TypeError:
```
jeanconn: This doesn't quite work as I think you'd hoped
```
"ValueError: could not convert string to float: '2019:162'
```
so I think we can just use ValueError instead of TypeError.

jeanconn:
I think I've addressed all comments.  The only execution difference was using a ValueError instead of TypeError as noted.

PR #330, opened 2019-07-12
Use kadi dynamic commanded states for starcheck thermal model

jeanconn:
Use kadi dynamic commanded states for starcheck thermal model

Closes #329

Supersedes #319

### DIFF
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -113,8 +113,8 @@ def get_ccd_temps(oflsdir, outdir='out',
     :param model_spec: xija ACA model specification
     :param run_start_time: Chandra.Time date used as a reference time to determine initial
                      seed state with temperature telemetry.  The initial seed state will
-                     be at the end of available telemetry that is also before run_start_time,
-                     before the beginning of backstop cmds, and before "now".
+                     be at the end of available telemetry that is also before run_start_time
+                     and before the beginning of backstop cmds.
     :param verbose: Verbosity (0=quiet, 1=normal, 2=debug)
     :returns: JSON dictionary of labeled dwell intervals with max temperatures
     """
@@ -158,10 +158,10 @@ def get_ccd_temps(oflsdir, outdir='out',
     proc['datestop'] = DateTime(tstop).date
 
     # Get temperature telemetry for 1 days prior to
-    # min(last available telem, backstop tstart, NOW, run_start_time)
-    # where run_start_time is basically a mock NOW for regression testing.
+    # min(last available telem, backstop tstart, run_start_time)
+    # where run_start_time is for regression testing.
     msid_range = fetch.get_time_range('aacccdpt', format='secs')
-    tlm = get_telem_values(min(msid_range[1], tstart, DateTime().secs, run_start_time.secs),
+    tlm = get_telem_values(min(msid_range[1], tstart, run_start_time.secs),
                            ['aacccdpt'],
                            days=1)
 
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -160,8 +160,8 @@ def get_ccd_temps(oflsdir, outdir='out',
     # Get temperature telemetry for 1 days prior to
     # min(last available telem, backstop tstart, run_start_time)
     # where run_start_time is for regression testing.
-    msid_range = fetch.get_time_range('aacccdpt', format='secs')
-    tlm = get_telem_values(min(msid_range[1], tstart, run_start_time.secs),
+    end_time = fetch.get_time_range('aacccdpt', format='secs')[1]
+    tlm = get_telem_values(min(end_time, tstart, run_start_time.secs),
                            ['aacccdpt'],
                            days=1)
 
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -161,7 +161,7 @@ def get_ccd_temps(oflsdir, outdir='out',
     # min(tstart, NOW, run_start_time) where run_start_time is basically
     # a mock NOW for regression testing.
     tlm = get_telem_values(min(tstart, tnow, run_start_time.secs),
-                           ['aacccdpt', 'pitch'],
+                           ['aacccdpt'],
                            days=30)
 
     states = get_week_states(tstart, tstop, bs_cmds, tlm)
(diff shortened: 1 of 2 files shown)


jeanconn:
@taldcroft I think this is all looking good, but what would you like to see as documentation of testing for both the PR and for LR? 

As we've changed
 - method of getting states for continuity interval
 - handling of eclipses (from none to some handling) 
 - determination of initial state with temperature from telemetry

I'm not sure if each needs to be validated independently or if "acq and guide temps for a set of schedules match expectations" is enough.

taldcroft:
Showing that "acq and guide temps for a set of schedules match expectations" is enough, where of course the set of schedules includes a couple with eclipses.

jeanconn:
Regarding eclipses, I suppose I can use the proseco pkl files for straightforward independent determination of those expectations.

taldcroft:
Or use one of the eclipse loads that failed starcheck but passed FOT thermal, and show now passing.  I would approve that.

starcheck/src/starcheck.pl
```diff
@@ -659,8 +679,7 @@ sub json_obsids{
 my %save_hash;
 
 my $out = '<TABLE><TD><PRE> ';
-my $date = `date`;
-chomp $date;
+my $date = date_for_printing($RUN_START_TIME);
```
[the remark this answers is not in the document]
jeanconn: Agreed and good point! (though I think we conflate happily in the ACIS checkers).  

I'd actually be inclined to skip adding anything new, as the thing we want, the seed time and temperature that is used for the thermal model, isn't available in this code.  I think we should just open a separate issue to actually save that out from calc_ccd_temps (because I don't think it is in either the return data from that tool or the saved-out json from the parent process.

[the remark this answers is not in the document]
taldcroft: Skip adding anything new sounds good.

jeanconn:
Regarding the states in general, I was happy to see the temperatures matched the MCC/pkl ones much better for the eclipse schedules, though I wasn't sure if we should add eclipse to the starcheck aca thermal model plot.

I also thought these were close to the mcc values but not quite *as* close as I would have guessed.

jun1019a_diff
jun1719a_diff
jun2419a_diff

taldcroft:
Re eclipse checking, looks great, I'll call that a pass!

starcheck/calc_ccd_temps.py
```diff
@@ -152,33 +150,32 @@ def get_ccd_temps(oflsdir, outdir='out',
...
-    # Get temperature telemetry for 30 days prior to
-    # min(tstart, NOW, run_start_time) where run_start_time is basically
-    # a mock NOW for regression testing.
-    tlm = get_telem_values(min(tstart, tnow, run_start_time.secs),
-                           ['aacccdpt', 'pitch'],
-                           days=30)
+    # Get temperature telemetry for 1 days prior to
+    # min(last available telem, backstop tstart, NOW, run_start_time)
+    # where run_start_time is basically a mock NOW for regression testing.
+    msid_range = fetch.get_time_range('aacccdpt', format='secs')
```
jeanconn: @taldcroft I think this is a real change relative to your kadi-states branch.  Do you want me to try to make this more readable?

taldcroft: I think you can cut NOW.  I don't see how telemetry can extend past NOW.  Otherwise OK.  I do generally prefer plural names for lists e.g. `msid_times`, or use `end_time = fetch.get_time_range(..)[1]`.

jeanconn: Good point about telemetry not extending into the future!

starcheck/src/starcheck.pl
```diff
@@ -137,7 +142,7 @@
     # time to be a time run_start_time days back from backstop start
     try:
         run_start_time = float(run_start_time)
-    except ValueError:
+    except (TypeError, ValueError):
```
jeanconn: This is really a bugfix on #329, so I'm not sure if I should just put it here or in a new PR.

taldcroft: Fine here.  Did this come up with `None`?

taldcroft:
Code looks good now, thanks!  Just need to re-run the validation plots and standard regression.

starcheck/src/starcheck.pl
jeanconn: Yes.  I probably should add a comment.  And I forgot that you might not git a github notification when I just commented on the other commit https://github.com/sot/starcheck/commit/dca234ef29dd98ed0cdf6bcae1bb9cce17317785#commitcomment-34330977

jeanconn:
Regarding standard regression, I ran with the "-3" run_start_time vs current master (not release) and get text diffs that look reasonable to me, with temperature differences that look up to 0.1 deg on first scroll-through:

[regress_diffs.txt](https://github.com/sot/starcheck/files/3401569/regress_diffs.txt)

jeanconn:
In more review it looks like there are some larger temperature diffs.  I notice some in JAN1215B which should have eclipses in it.  I would prefer to do more rigorous eval of the diffs (and confirm that all larger diffs are due to eclipses) but I think basically we are OK and any reasonable diffs related to temperature are fine. We're mostly using the full regression set to just confirm that everything runs and there are no new weird errors.

starcheck/src/starcheck.pl
jeanconn: I added a comment.

jeanconn:
For the validation plots, I ran these with these start times
```
./sandbox_starcheck -dir kadistates/jun1019a/ -out jun1019a_kadi -run_start_time 2019:155:14:00:00.000
./sandbox_starcheck -dir kadistates/jun1719a/ -out jun1719a_kadi -run_start_time 2019:162:22:00:00.000
./sandbox_starcheck -dir kadistates/jun2419a/ -out jun2419a_kadi -run_start_time 2019:169:00:00:00.000

./sandbox_starcheck -dir kadistates/mar2519a/ -out mar2519a_kadi -run_start_time '2019:079:22:00:00.000'
./sandbox_starcheck -dir kadistates/mar1119b/ -out mar1119b_kadi -run_start_time '2019:065:18:00:00.000'
```
And rerun of the weeks and their plots comparing to the pickle temperatures look the same

jun1019a_diff
jun1719a_diff
jun2419a_diff
mar1119b_diff
mar2519a_diff

jeanconn:
@taldcroft Is this officially good to go now?

PR #334, opened 2019-08-06
Switch to using proseco agasc file as default.

jeanconn:
Switch to using proseco agasc file as default.

### DIFF
--- a/starcheck/src/starcheck.pl
+++ b/starcheck/src/starcheck.pl
@@ -171,7 +171,7 @@ my %par = (dir  => '.',
 		   html => 1,
 		   text => 1,
 		   yaml => 1,
-                   agasc_file => "${SKA}/data/agasc/agasc1p7.h5",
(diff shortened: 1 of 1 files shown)


jeanconn:
I'll need to think a bit more on what we might want to do to show this is fine / safe .

taldcroft:
What's the driver?

jeanconn:
Right now, we're using all three (agasc1p7.h5, miniagasc1p7.h5, and proseco_agasc_1p7.h5) inst starcheck.

I'd like to cut down the dependencies from a maintenance perspective; and also I think in the agasc module aren't we building a Dec index for each of these as they are loaded?  So I think it isn't super memory efficient either.

I just noticed it again when working standalone and I thought I had everything I wanted by my ska_sync didn't have the bigger agasc file.

But mostly I put the PR here because I thought it was a good idea for one minute and wanted to capture it.  Didn't necessarily want to push this now.

taldcroft:
Sounds good.  I see no problem in principle with using proseco_agasc everywhere (including in place of miniagasc).

PR #335, opened 2019-08-06
Add checks for skare3 perl and deps in launchers

jeanconn:
Add checks for skare3 perl and deps in launchers

This will print more useful output if the conda environment doesn't have the necessary Perl pieces to run starcheck for load review.

There's still duplication between `sandbox_starcheck` and `starcheck` but overall I think this is still benign.

### DIFF
--- a/sandbox_starcheck
+++ b/sandbox_starcheck
@@ -1,10 +1,28 @@
 #!/bin/bash
 
 DIR=`dirname $0`
+export PYTHONPATH=$DIR:$PYTHONPATH
+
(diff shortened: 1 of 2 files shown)


jeanconn:
@taldcroft Do you need me to reorganize to avoid the DRY here or is this passable from your perspective?

jeanconn:
It was also the plan to make these tiny tweaks to the starcheck launchers as part of reorganization for https://github.com/sot/mica/issues/204

PR #336, opened 2019-08-22
Update characteristics for new -8.8 and -7.8 limits

jeanconn:
Update characteristics for new -8.8 and -7.8 limits


jeanconn:
Not sure if it is worth doing the nomenclature changes here.

taldcroft:
It's true that temperature red limit and yellow limit have generally accepted meanings in operations that don't match how they are used in starcheck.  But it's fine.

PR #337, opened 2019-08-23
Remove sybase use for star stats

jeanconn:
Remove sybase use for star stats

This is a redo of #257 and #187

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,6 @@ FLIGHT_ENV = SKA
 
 SRC = starcheck/src
 
-#include $(SKA)/include/Makefile.FLIGHT
 
 RELATED_LIB = $(SRC)/StarcheckParser.pm
 BIN = $(SRC)/starcheck.pl $(SRC)/starcheck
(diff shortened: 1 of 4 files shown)


jeanconn:
Regarding

https://github.com/sot/starcheck/pull/187#discussion_r179206680

I did compare this current approach (fetching all acq and guide stats and then using non-indexed searches within them) to using the agasc_id indexes on the hdf5 tables (via tables.read_where).  

For an average set of stars in a schedule, it looks like using the indexed approach would reduce this part of the runtime by a bit more than a third (2.2 seconds to use the pre-fetch, 1.4 seconds to open the hdf5, keep it open, and do indexed queries).  The 0.8 seconds did not seem worth it in this case (though fetching all the stats to start is not particularly memory efficient either).

jeanconn:
@taldcroft Do you want this to go too or would you like more time to consider?

jeanconn:
IIRC `acqs` and `guides` were put into `Table` from this suggestion https://github.com/sot/starcheck/pull/257#discussion_r224488994

jeanconn:
I can add a comment about that if it isn't going to be the convention to just `Table` things if the code holds still long enough.

taldcroft:
Thanks for refreshing my memory.  Indeed any data coming straight from HDF5 should generally be `Table`'d so that character data can be handled sensibly.  So to answer my question, the string comparison `ok = acqs['img_func'] == 'star'` does require `Table`.

PR #341, opened 2020-04-18
Use RLTT and SCHEDULE_STOP_TIME in thermal calc

### LINKED ISSUE #340: Use RUNNING_LOAD_TERMINATION_TIME for ACA thermal propagation
I believe that the following lines are the key ones for taking advantage of `RUNNING_LOAD_TERMINATION_TIME` and `SCHEDULED_STOP_TIME` commands if available.  The idea is to override `tstart` and `tstop` with those command times.

https://github.com/sot/starcheck/blob/74cac71763f88deb143bd179bd343eb988332b6b/starcheck/calc_ccd_temps.py#L155

At the point below, with the new version of `tstart` it is possible that `cmds` is not time-ordered, so that needs to be taken care of.  (This is a consequence of the fact that schedule commands can overlap.)

https://github.com/sot/starcheck/blob/74cac71763f88deb143bd179bd343eb988332b6b/starcheck/calc_ccd_temps.py#L313

jeanconn:
## Description

Use RLTT and SCHEDULE_STOP_TIME in thermal calc

## Testing

- [ ] Passes unit tests on MacOS, linux, Windows (at least one required)
- [ ] Functional testing

Fixes #340

### DIFF
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -152,8 +152,19 @@ def get_ccd_temps(oflsdir, outdir='out',
 
     # Get tstart, tstop, commands from backstop file in opt.oflsdir
     bs_cmds = get_bs_cmds(oflsdir)
-    tstart = DateTime(bs_cmds[0]['date']).secs
-    tstop = DateTime(bs_cmds[-1]['date']).secs
+
+    # Use RLTT and SCHEDULED_STOP_TIME if available
+    rltt = bs_cmds['event_type'] == 'RUNNING_LOAD_TERMINATION_TIME'
+    if np.any(rltt):
+        tstart = DateTime(bs_cmds['date'][rltt][0]).secs
+    else:
+        tstart = DateTime(bs_cmds[0]['date']).secs
+    sched_stop = bs_cmds['event_type'] == 'SCHEDULED_STOP_TIME'
+    if np.any(sched_stop):
+        tstop = DateTime(bs_cmds['date'][sched_stop][0]).secs
+    else:
+        tstop = DateTime(bs_cmds[-1]['date']).secs
+
     proc['datestart'] = DateTime(tstart).date
     proc['datestop'] = DateTime(tstop).date
 
@@ -294,8 +305,8 @@ def get_week_states(tstart, tstop, bs_cmds, tlm):
     """
     Make states from last available telemetry through the end of the backstop commands
 
-    :param tstart: start time from first backstop command
-    :param tstop: stop time from last backstop command
+    :param tstart: start time from RLTT if available else first backstop command
+    :param tstop: stop time from SCHEDULED_STOP_TIME if available else last backstop command
     :param bs_cmds: backstop commands for products under review
     :param tlm: available pitch and aacccdpt telemetry recarray from fetch
     :returns: numpy recarray of states
@@ -306,7 +317,7 @@ def get_week_states(tstart, tstop, bs_cmds, tlm):
     init_tlm_time = np.mean(tlm['time'][ok])
 
     # Get commands from last telemetry up to (but not including)
-    # first backstop command.
+    # commands at tstart (RLTT if present else first backstop cmd)
     cmds = kadi.commands.get_cmds(init_tlm_time, tstart)
 
     # Add in the backstop commands
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -154,16 +154,20 @@ def get_ccd_temps(oflsdir, outdir='out',
     bs_cmds = get_bs_cmds(oflsdir)
 
     # Use RLTT and SCHEDULED_STOP_TIME if available
-    rltt = bs_cmds['event_type'] == 'RUNNING_LOAD_TERMINATION_TIME'
-    if np.any(rltt):
-        tstart = DateTime(bs_cmds['date'][rltt][0]).secs
+    ok = bs_cmds['event_type'] == 'RUNNING_LOAD_TERMINATION_TIME'
+    if np.any(ok):
+
+        # The RLTT is defined so that running commands at exactly the RLTT are
+        # included in propagation. However get_cmds() uses the convention of
+        # getting commands date_start <= cmd_date < date_stop, so add 0.001 to RLTT.
+        tstart = DateTime(bs_cmds['date'][ok][0]).secs + 0.001
     else:
-        tstart = DateTime(bs_cmds[0]['date']).secs
-    sched_stop = bs_cmds['event_type'] == 'SCHEDULED_STOP_TIME'
-    if np.any(sched_stop):
-        tstop = DateTime(bs_cmds['date'][sched_stop][0]).secs
+        tstart = DateTime(bs_cmds['date'][0]).secs
+    ok = bs_cmds['event_type'] == 'SCHEDULED_STOP_TIME'
+    if np.any(ok):
+        tstop = DateTime(bs_cmds['date'][ok][0]).secs
     else:
-        tstop = DateTime(bs_cmds[-1]['date']).secs
+        tstop = DateTime(bs_cmds['date'][0]).secs
 
     proc['datestart'] = DateTime(tstart).date
     proc['datestop'] = DateTime(tstop).date


starcheck/calc_ccd_temps.py
```diff
@@ -152,8 +152,19 @@ def get_ccd_temps(oflsdir, outdir='out',
 
     # Get tstart, tstop, commands from backstop file in opt.oflsdir
     bs_cmds = get_bs_cmds(oflsdir)
-    tstart = DateTime(bs_cmds[0]['date']).secs
-    tstop = DateTime(bs_cmds[-1]['date']).secs
+
+    # Use RLTT and SCHEDULED_STOP_TIME if available
+    rltt = bs_cmds['event_type'] == 'RUNNING_LOAD_TERMINATION_TIME'
```
taldcroft, line 157: Always use `ok` per Ska idioms in this situation.  This tells the reader right away that it is a temporary selection mask.

starcheck/calc_ccd_temps.py
```diff
@@ -152,8 +152,19 @@ def get_ccd_temps(oflsdir, outdir='out',
...
+    if np.any(rltt):
+        tstart = DateTime(bs_cmds['date'][rltt][0]).secs
+    else:
+        tstart = DateTime(bs_cmds[0]['date']).secs
```
taldcroft, line 161: Make this consistent with the line above with `bs_cmds['date'][0]`.  This brings it in lines with current best practice of always selecting the column first, then the item within the column.  This is much faster.  Yes, it is changing code, but in a way that is safe and leaves this code in a better way.

starcheck/calc_ccd_temps.py
```diff
@@ -294,8 +305,8 @@ def get_week_states(tstart, tstop, bs_cmds, tlm):
     """
     Make states from last available telemetry through the end of the backstop commands
 
-    :param tstart: start time from first backstop command
-    :param tstop: stop time from last backstop command
+    :param tstart: start time from RLTT if available else first backstop command
```
taldcroft, line 308: This arg name is now quite misleading since it has nothing to do with the "start" of anything.  It is really the "end" of something, namely the end of running loads.  So I'd suggest changing it to `t_rltt`.  I assume this function is not called directly from anywhere outside of this same module.

starcheck/calc_ccd_temps.py
```diff
@@ -294,8 +305,8 @@ def get_week_states(tstart, tstop, bs_cmds, tlm):
...
+    :param tstop: stop time from SCHEDULED_STOP_TIME if available else last backstop command
```
taldcroft, line 309: Give the description `schedule stop time from ...`

starcheck/calc_ccd_temps.py
```diff
@@ -152,8 +152,19 @@ def get_ccd_temps(oflsdir, outdir='out',
 
     # Get tstart, tstop, commands from backstop file in opt.oflsdir
     bs_cmds = get_bs_cmds(oflsdir)
-    tstart = DateTime(bs_cmds[0]['date']).secs
-    tstop = DateTime(bs_cmds[-1]['date']).secs
+
+    # Use RLTT and SCHEDULED_STOP_TIME if available
+    rltt = bs_cmds['event_type'] == 'RUNNING_LOAD_TERMINATION_TIME'
+    if np.any(rltt):
+        tstart = DateTime(bs_cmds['date'][rltt][0]).secs
```
taldcroft, line 159: This needs to be `tstart = DateTime(bs_cmds['date'][rltt][0]).secs + 0.001`.  The RLTT is defined so that running commands at exactly the RLTT are *included* in propagation.  However `get_cmds()` uses the convention of getting commands `date_start <= cmd_date < date_stop`.  (All done using standard date strings to 1 millisec precision).

https://cxc.cfa.harvard.edu/mta/ASPECT/tool_doc/kadi/api.html#kadi.commands.commands.get_cmds

jeanconn: Gotcha.  I still hadn't understood the plan but figured just tossing it in here would help.  I'd be helped more by some worked examples in test cases.

taldcroft: The recent interrupt schedule had an example of this.  But maybe you are overthinking this since it is really no more than "chop running commands after RLTT, propagate out to scheduled stop".  That's it.  

All the words John sent about how MP chooses the times are not relevant except for FOT MP is guaranteeing that "chop running commands after RLTT, propagate out to scheduled stop" will give the correct set of commands for propagation.

jeanconn: "The recent interrupt schedule had an example of this."  I thought the schedule was incorrect about RLTT and they were going to change for future use.  And I didn't have the "right answer".  But that's fine.

taldcroft: Indeed the RLTT for APR1420 would include an AOACRSTD command that never ran.  Though from the perspective of "giving the correct set of commands for propagation", it was just fine because AOACRSTD doesn't affect propagation.  In fact that more functional definition (giving correct propagation) is really applicable because in the case of an interrupt there might be transmitter off commands before the RLTT that never actually get run, but we don't care.

jeanconn: Also, why did the AOACRSTD not run?  I thought the RLTT in the APR1420 backstop was really describing the time to terminate the APR1320 commands running onboard.  The AOACRSTD command was in the APR1420 backstop.  Does it only run contingent upon something else?

taldcroft: I was referring to the AOACRSTD at 2020:105:01:05:22.340 in APR1320 that was being interrupted since we have "Estimated time of on-board load stop time: 2020:104:22:25:00.000" (referring to the real-time activity of stopping loads).

The way I said it was admittedly ambiguous in language.  How about "Indeed the RLTT specified in the APR1420 loads would specify to include (in the propagation commands) an AOACRSTD command from APR1320A that occurs well after the actual interrupt time during comm."  But this is because of a one-time process mistake and an off-by-1msec error.  Going forward that won't happen.

Takeaway: the RLTT is not *exactly* related to when loads will actually be interrupted by operators during comm.  It will normally be after the comm, with the only hard requirement that chopping commands using the RLTT will give the correct propagation.

jeanconn: Ah.  I didn't realize that there were *two* commands at exactly the RLTT for this interrupt, the AOACRCSTD at 2020:105:01:05:22.340 in the interrupted and interrupting schedules.  I assume the AOACRCSTD from APR1420A did run (and it doesn't really matter).  Your comment here about RLTT not necessarily being when the loads will actually be interrupted is also helpful to me (not for this starcheck exercise but in general).

starcheck/calc_ccd_temps.py
jeanconn: Well, it is the start time of the interval over which we are getting week states.

taldcroft: The interval for week states is `(init_tlm_time, schedule_stop_time)`, so `tstart` is somewhere in the middle.  All this stuff is confusing, so it is really important to be clear with variable names and comments.

starcheck/calc_ccd_temps.py
taldcroft: The name should be `scheduled_stop`.  It doesn't need to be in seconds (tstop) at this point since it will be required as a date string later.

taldcroft: Note that `tstop` is not actually even being used right now.  It needs to be used to set the `datestop` of the last state down below.  Something like:
```
states['datestop'][-1] = scheduled_stop
# Right after the get_states() call`
```

starcheck/calc_ccd_temps.py
```diff
@@ -306,7 +321,7 @@ def get_week_states(tstart, tstop, bs_cmds, tlm):
     init_tlm_time = np.mean(tlm['time'][ok])
 
     # Get commands from last telemetry up to (but not including)
-    # first backstop command.
+    # commands at tstart (RLTT if present else first backstop cmd)
     cmds = kadi.commands.get_cmds(init_tlm_time, tstart)
```
taldcroft, line 325: With https://github.com/sot/kadi/pull/163 this should be:
```
cmds = kadi.commands.get_cmds(init_tlm_time, rltt, inclusive_stop=True)
```

taldcroft:
Superceded by #344, which is nearly approved.

PR #347, opened 2020-07-01
Check for perl based on location of Ska Python

taldcroft:
## Description

This is basically what is used for the sandbox starcheck, and it appears to work fine.

## Testing

- [N/A] Passes unit tests
- [x] Functional testing (partial)

### Minimal testing

I installed this into a ska3-shiny environment on Mac and confirmed that it ran without error on a representative recent set of load products.

### DIFF
--- a/starcheck/src/starcheck
+++ b/starcheck/src/starcheck
@@ -8,12 +8,15 @@
 # Unset PYTHONPATH because custom PYTHONPATH should never be allowed
 # for flight starcheck
 unset PYTHONPATH
-export PYTHONHOME=$SKA
+
+PYTHON_EXE=`which python`
+PYTHON_DIR=`dirname $PYTHON_EXE`
+export PYTHONHOME=`dirname $PYTHON_DIR`
 
 # Check for skare3 perl
 if [[ ! -f ${PYTHONHOME}/bin/perl ]];
 then
-    echo "skare3 perl not installed. conda install perl perl-core-deps perl-ska-classic"
+    echo "skare3 perl not installed. conda install perl perl-ska-convert"
     exit 1
 fi
 # Check for perl deps


jeanconn:
Still would need testing on ska3-flight to go in `master` but besides that fine.

jeanconn:
This looks to work for me in a root linux test ska.  The only perfect testing issue is that is still a bit different from the call to /proj/sot/ska/bin/starcheck that calls /proj/sot/ska3/flight/bin/starcheck.  I'm good with the change.

starcheck/src/starcheck
```diff
@@ -8,12 +8,15 @@
...
+
+PYTHON_EXE=`which python`
+PYTHON_DIR=`dirname $PYTHON_EXE`
+export PYTHONHOME=`dirname $PYTHON_DIR`
 
 # Check for skare3 perl
 if [[ ! -f ${PYTHONHOME}/bin/perl ]];
 then
-    echo "skare3 perl not installed. conda install perl perl-core-deps perl-ska-classic"
+    echo "skare3 perl not installed. conda install perl perl-ska-convert"
```
jeanconn, line 19: It probably makes more sense to just make a metapackage that has the perl-starcheck-needs, but this change should be fine for now even if it isn't consistent with the similar printent statement a few lines below.

taldcroft:
Should we put this into a `shiny` branch to punt off needing to test twice (in current ska3-flight and shiny?).  That's kind of the way I've been going with changes like this.

jeanconn:
For the launcher, I've already done the integration testing that is required for master promotion, so this can just go in master now.  Because of the nature of the launcher, I still note that this is a change that should be checked after promotion (as in, this kind of change should get a "SOT MP should run this after promotion" test).  No big deal.  I would like to get a few more changes in for starcheck soon and haven't thought of anything that really needs a shiny branch.

PR #348, opened 2020-07-05
Trim Perl dependencies by moving more to Python

jeanconn:
## Description

Trim Perl dependencies by moving more to Python.

Specifically, this removes the rather heavy PDL dependency and also cuts the various time-related dependencies.
It also seemed time to move a bunch of Python into starcheck.utils to make it easier to call from multiple pieces of the Perl.

It also just removes the Dark Cal Checker.

## Testing

~~Passes unit tests on MacOS, linux, Windows (at least one required)~~ No unit tests
- [x] Functional testing - ran sandbox_starcheck on several recent loads from Linux without errors
- [x] Regression testing - ran on Linux and see only expected changes (all dark cal reporting is gone)


jeanconn:
My thought regarding the Dark Cal checker is that we can either run an old version of starcheck or resurrect the Perl module if we go back to doing those activities.

taldcroft:
> My thought regarding the Dark Cal checker is that we can either run an old version of starcheck in a pinch or resurrect the Perl module if we go back to doing those activities.

I'm good with cutting the dark cal checker.  It is extremely unlikely we will go back to doing cals that way.

jeanconn:
Great!  And yeah, I figured it was fine to cut but figured I'd talk it out in PR just like we have.

starcheck/src/starcheck.pl
```diff
@@ -50,107 +46,15 @@
...
-    except (TypeError, ValueError):
-        ref_time = DateTime(run_start_time)
-    else:
-        if run_start_time < 0:
-            ref_time = DateTime(backstop_start) + run_start_time
-        else:
-            raise ValueError("Float run_start_time should be negative")
-    return ref_time.date
+from starcheck.pcad_att_check import check_characteristics_date
+from starcheck.utils import (_make_pcad_attitude_check_report,
```
taldcroft, line 51: YES!!!!  Excellent idea.

starcheck/src/lib/Ska/Parse_CM_File.pm
```diff
@@ -479,8 +487,8 @@ sub DOT {
 
     foreach (keys %command) {
         %{$dot{$_}} = parse_params($command{$_});
-        $dot{$_}{time}  = date2time($dot{$_}{TIME})     if ($dot{$_}{TIME});
-        $dot{$_}{time} += date2time($dot{$_}{MANSTART}) if ($dot{$_}{TIME} && $dot{$_}{MANSTART});
+        $dot{$_}{time}  = date2time($dot{$_}{TIME}) if ($dot{$_}{TIME});
+        $dot{$_}{time} += rel_date2time($dot{$_}{MANSTART}) if ($dot{$_}{TIME} && $dot{$_}{MANSTART});
```
jeanconn, line 491: I should add a comment here about why the MANSTART time needs the relative datetime method, and should add a comment to the newish rel_date2time method indicating what time formats it expects.

starcheck/src/starcheck.pl
```diff
@@ -346,22 +250,6 @@
     warning("uname == SunOS; starcheck is only approved on Linux \n");
```
jeanconn, line 250: I can just cut up a few more lines from the Dark cal checker and remove this too.

jeanconn:
Terrific.  What would you like to see for functional testing?  And indeed I should have documented.  Right now I've run on recent test loads and it runs in ska3 and makes output.  Since making the PR, I've also run the old complete regression test set and they run without issue (functional component) and the output looks good with the only diffs being that all the dark cal checker output is correctly missing (regression tests pass).

starcheck/src/starcheck.pl
jeanconn: Well, it was your idea and it was a good one!  So much nicer to keep the Python in a separate file and just convenient for editing.  I think some of the exception handling and wrappers still need simplification and improvement, but moving them into the separate file is a good start.

taldcroft:
Sounds good.

PR #327, opened 2019-07-01
WIP: Call sparkles from starcheck

jeanconn:
Call sparkles from starcheck

This was just weekend fun (and tool a lot longer than I'd hoped with many many Segmentation faults), but my thought was that it could be a simple way to mostly calling one set of checks on catalogs.  The next steps in this PR would be to cut out the starcheck checks that duplicate sparkles checks.

PR #356, opened 2020-10-07
Remove orphaned scripts / makefile

taldcroft:
## Description

Neither of the two scripts are called anywhere and bad_stars.py apparently hasn't been run in 5 years. This is all overtaken by improvements in mica (bad obsid list and mag stats). The bad_stars script relies on an unmaintained sybase table.

The Makefile just seems a leftover from another era, but not 100% sure.

This came up in working on removing Sybase dependencies.

## Testing

- [N/A] Passes unit tests : no installed code calls either of these scripts
- [N/A] Functional testing


jeanconn:
Oh, and I have been using the top level Makefile not the one in starcheck/data/Makefile.  The one you suggest removing seems fine to go.

jeanconn:
Since we are still using the agasc.bad list made with bad_stars.py I don't know if we want to remove the script (and manual_bad_stars) until we swap to using the agasc supplement (which I suppose we could do at any point, I was just headed to doing it more via #327 ).

taldcroft:
I think of `agasc.bad` just being a fixed legacy file that we will continue to use for now but never again update. The `bad_stars.py` has not been run since 2015 to update that file, so I really don't see why not to just start tidying up a bit.

taldcroft:
The driver of course being that `bad_stars.py` uses sybase and gives the appearance of a sybase dependency.

jeanconn:
I already removed agasc.bad in the sparkles branch I think, because I was letting sparkles do the bad star tests from the agasc.  My point about maybe waiting and removing together was that bad_stars.py at least can tell you why a star is in the current list.

taldcroft:
I really don't think we care at this point, and it's going to all be overtaken by mag-stats.

jeanconn:
OK.  Then this PR could remove the manual stars file and the sausage one too.

taldcroft:
Done, removed those two.

PR #358, opened 2020-11-05
Fix info statement regarding being off padded CCD

taldcroft:
## Description

The statement that P_ACQ should be less than 0.5 was misleading.

## Testing

No testing done since this is just changing the text of a message.

- [N/A] Passes unit tests on MacOS, linux, Windows (at least one required)
- [N/A] Functional testing


jeanconn:
This change doesn't bother me.  I was planning to chuck it altogether in #327.  I suppose I was also hesitant to cut a starcheck release before shiny with the thought that briefing any changes to load review tools in the context of shiny promotion was not ideal.

taldcroft:
I don't care if this gets merged or promoted any time soon, I just wanted to do the PR while it was on my mind.

jeanconn:
OK.  Then I'm good to merge it now.

PR #362, opened 2021-01-10
Remove special case ER checks

jeanconn:
## Description

Remove special case ER checks

I've left this in the load review checklist, so if it comes up again we can revisit using those rules.

## Testing

- [x] Passes unit tests on MacOS, linux, Windows (at least one required)
   - No unit tests
- [x] Functional testing
   - Ran on a recent schedule without erroring out

Fixes #

### DIFF
--- a/starcheck/src/aca_load_review_cl.rst
+++ b/starcheck/src/aca_load_review_cl.rst
@@ -208,22 +208,6 @@ Checks
 |       |                  | | |configuration (2    |      |     |    |    |    |                |
 |       |                  | | |monitor windows)    |      |     |    |    |    |                |
 +-------+------------------+-+-+--------------------+------+-----+----+----+----+----------------+
-|       |                  |X|X|Special Case Engineering Request           |    |                |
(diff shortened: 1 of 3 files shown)


taldcroft:
You should remove the same from the review checklist since we are no longer checking this. If something drastic changes then we revert this PR to put it back.

jeanconn:
OK.  I figured not removing the checklist section was one option, as not checking for special case ers means not relaxing a set of criteria, and we'd have the checklist handy to review one manually if it got all red warnings.  Removing the checklist item I suppose puts this squarely in requiring SSAWG review, but it was getting there anyway.

taldcroft:
The special-case ER was an artifact of the OFLS scheduler. We won't ever use that again, so the thing which is known as a special-case ER will never happen. There might be something else that requires a new rule, but it won't be that.

jeanconn:
OK.  I've removed the checklist block.  I thought there could still be a reason to run an ER with "relaxed" star field constraints if it is running outside of perigee, but if it is just not going to come up that is fine.

taldcroft (approved):

PR #367, opened 2021-01-10
Remove old ok-no-starcat stuff

jeanconn:
## Description

Remove old ok-no-starcat stuff

This code was meant for SIR ERs or dark cals, and isn't working on modern undercovers and other anyway.  Might as well just get a bunch of red warnings on an undercover etc and remove the cruft.

## Testing

- [x] Passes unit tests on MacOS, linux, Windows (at least one required)
   - No unit tests
- [x] Functional testing
   - Confirmed an old dark cal week (DEC1115A) just shows red warns for the dark cals

Fixes #

### DIFF
--- a/starcheck/data/characteristics.yaml
+++ b/starcheck/data/characteristics.yaml
@@ -1,14 +1,6 @@
 paths:
    week_lookup: https://icxc.harvard.edu/cgi-bin/aspect/starcheck/find_shortterm.cgi
 
-# string match on oflsids that may not have a maneuver or starcat, and its OK!
-# The RDE (and probably RDX) entries most likely do not applie to ORViewer DOT schedules
-# and can likely be removed in the future.
-no_starcat_oflsid:
-   - RDX
-   - RDE
-   - DC_T
-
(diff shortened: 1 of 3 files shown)


taldcroft:
I haven't looked at the code changes in detail, but I'm a little worried on principle. It sounds like this is going to introduce spurious red warnings where the code is currently suppressing them? Maybe I have that wrong, but I hope we can (to the extent practical) avoid issuing red warnings that we "just know" to ignore.

jeanconn:
Not quite.  The code is already issuing somewhat spurious red warnings (on the 62??? part of an undercover).  This code supported a special set of OFLSids we no longer use to stop red warnings on SIR (IIRC) and dark cals.

If we wanted to try to cut the red warnings from undercovers and gyro holds based on OFLSid, we could extend this code, but in general, I think it was and is a bad idea to use the oflsids this way.

taldcroft (approved):
OK, I looked through this and it seems the code changes correspond to always having `ok_no_starcat` be undefined or False. That seems uncontroversial.

PR #372, opened 2021-04-13
Date and increment version of checklist

jeanconn:
## Description

Date and increment version of checklist to 3.7

## Testing

No testing - does not apply.


jeanconn:
The checklist is installed outside of the starcheck package process, so this does not need to be in the starcheck release for ska3-flight 2021.4, but I think the version should be captured in main.

PR #373, opened 2021-06-24
Add suspect bad pixel region to bad pixels

jeanconn:
## Description

Add bad pixel region to check for stars too close to the CCD region in which diminished star mags are apparent when star dithers through. See slack aspect-team conversation Jun 25, 2021 (search "composite image of the bad region for all observations since 2016").

## Testing

-  [x] Functional testing

Ran on MAR3020A and JUL0719A and confirmed obsids 22957 and 48040 are marked with CRITICAL warnings with stars near a bad pixel.


jeanconn:
I'd prefer to do this as part of #327 , but given the possible impact, we might want to just toss this in as an interim check.

jeanconn:
Additionally, I don't have a strong signal for the extent of what I'm calling a bad pixel region, but these numbers seem in the ballpark.

taldcroft (approved):
Looks good, testing looks good!

jeanconn:
Great.  Thanks.  This update is, as you saw, super trivial to get this protection in.

taldcroft:
Now we can at least run a dev starcheck on the side for testing.

PR #378, opened 2021-11-12
round guide count and P2

javierggt:
## Description

This PR adds changes to round P2 and guide count so we do not get warnings due to numerical errors (like we got in [NOV1521A](https://icxc.harvard.edu/mp/mplogs/2021/NOV1521/oflsa/starcheck.html).

## Testing

- [x] Functional testing. I ran this version on the [NOV1521A](https://icxc.harvard.edu/mp/mplogs/2021/NOV1521/oflsa/starcheck.html) loads. The result is [here](https://icxc.cfa.harvard.edu/aspect/tmp/starcheck-pr378-v3/starcheck.html), and the diff between the two starcheck files is [here](https://icxc.cfa.harvard.edu/aspect/tmp/starcheck-pr378-v3/diff.html)

### DIFF
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -2619,6 +2619,19 @@ sub quat2radecroll {
     return ($ra, $dec, $roll);
 }
 
+###################################################################################
+sub check_guide_count{
+###################################################################################
+    my $self = shift;
+    my $guide_count = $self->count_guide_stars();
+
+    if ($guide_count < 4.0){
+        push @{$self->{warn}}, "Guide count of $guide_count < 4.0.";
+    }
+
+    # Also save the guide count in the figure_of_merit
+    $self->{figure_of_merit}->{guide_count} = $guide_count;
+}
 
 ###################################################################################
 sub count_guide_stars{
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -2630,8 +2630,10 @@ sub check_guide_count{
     my $self = shift;
     my $guide_count = $self->count_guide_stars();
 
-    if ($guide_count < 4.0){
-        push @{$self->{warn}}, "Guide count of $guide_count < 4.0.";
+    my $min_num_gui = ($self->{obsid} >= 38000 ) ? 6.0 : 4.0;
+
+    if ($guide_count < $min_num_gui){
+        push @{$self->{warn}}, "Guide count of $guide_count < $min_num_gui.";
     }
 
     # Also save the guide count in the figure_of_merit
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -1120,6 +1120,7 @@ sub check_bright_perigee{
 	push @{$self->{warn}}, "$bright_count star(s) brighter than scaled 9th mag. "
 	    . "Perigee requires at least $min_n_stars\n";
     }
+    $self->{figure_of_merit}->{guide_count_9th} = $bright_count;
 }
 
 
@@ -2272,9 +2273,12 @@ sub print_report {
         $o .= sprintf("%.1f", $self->{figure_of_merit}->{P2}) . "\t";
 	$o .= "$font_stop" if $bad_FOM;
 	$o .= "\n";
-	$o .= sprintf("Acquisition Stars Expected  : %.2f\tGuide star count: %.1f \n",
-                      $self->{figure_of_merit}->{expected},
-                      $self->{figure_of_merit}->{guide_count});
+	$o .= sprintf("Acquisition Stars Expected  : %.2f\n", $self->{figure_of_merit}->{expected});
+	$o .= sprintf("Guide star count: %.1f \t", $self->{figure_of_merit}->{guide_count});
+	if (defined $self->{figure_of_merit}->{guide_count_9th}){
+	    $o .= sprintf("Guide count_9th: %.1f", $self->{figure_of_merit}->{guide_count_9th});
+	}
+	$o .= "\n";
     }
 
 
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -2457,7 +2457,7 @@ sub identify_stars {
             my $gs_mag = $c->{"GS_MAG$i"};
             my $dmag = abs($star->{mag_aca} - $gs_mag);
             if ($dmag > 0.01){
-                push @{$self->{warn}},
+                push @{$self->{yellow_warn}},
                     sprintf("[%d] Guide sum mag diff from agasc mag %9.5f\n", $i, $dmag);
             }
 	    # let's still confirm that the backstop yag zag is what we expect
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -2269,7 +2269,7 @@ sub print_report {
 	my $bad_FOM = $self->{figure_of_merit}->{cum_prob_bad};
 	$o .= "$red_font_start" if $bad_FOM;
 	$o .= "Probability of acquiring 2 or fewer stars (10^-x):\t";
-        $o .= substr(sprintf("%.4f", $self->{figure_of_merit}->{P2}), 0, 6) . "\t";
+        $o .= substr(sprintf("%.1f", $self->{figure_of_merit}->{P2}), 0, 3) . "\t";
 	$o .= "$font_stop" if $bad_FOM;
 	$o .= "\n";
 	$o .= sprintf("Acquisition Stars Expected  : %.2f\n",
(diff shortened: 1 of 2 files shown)


jeanconn (commented):
Looks good to me.  _guide_count looks to be explicitly returning a float, and the P2 from proseco_probs is explicitly a float, so the sprintf behavior should round it.  In the absence of unit tests I generally just try to find weeks that show what we want for functional tests.

jeanconn:
So the NOV1521A functional test shows that the bright guide count is working.  I figure we probably also want functional test coverage of the OR guide count and P2, or could let them go...

javierggt:
the OR guide count is already rounded in the output, it seems, so there is no associated diff.

jeanconn:
Right, and for the P2 it seems like we should probably update the format strong for the print as well (unless the zeros are now reassuring).

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -2269,7 +2269,7 @@ sub print_report {
 	my $bad_FOM = $self->{figure_of_merit}->{cum_prob_bad};
 	$o .= "$red_font_start" if $bad_FOM;
 	$o .= "Probability of acquiring 2 or fewer stars (10^-x):\t";
-        $o .= substr(sprintf("%.4f", $self->{figure_of_merit}->{P2}), 0, 6) . "\t";
+        $o .= substr(sprintf("%.1f", $self->{figure_of_merit}->{P2}), 0, 3) . "\t";
```
taldcroft, line 2272: What is the `substr` for here? I see the logic in the change but the original code is confusing me.

taldcroft (commented):
Overall the use of `sprintf` to do rounding and turn the value into a string really seem horrible. But this is perl and after a bit of googling that does seem to be the idiom. So 👍 on this. The question is just about functional testing and interface changes.

And speaking of which, is this a good time to include the guide count in the starcheck per-obsid report instead of just in the summary up top?.

javierggt:
I also do not like the use of `sprintf`. I could do `int(10*$number + 0.5)/10` or something like that, but I also googled... so...

starcheck/src/lib/Ska/Starcheck/Obsid.pm
javierggt: I am betting this can be removed.

I can imagine something like this to truncate the number if the rounded number takes more digits than expected, but it is also not a very good idea.

jeanconn:
Agreed.  And I commented on the inputs being floats because I seemed to remember that was the "gotcha" with using sprintf this way (I don't think it rounds if they are already strings)... but I didn't test.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
javierggt: this could related to sprintf actually rounding a float and not giving a string as output. When you round, you might not get a number with exact binary representation, and then when you print you get 0.9000001 instead of 0.9, so you truncate the string.

taldcroft:
> (I don't think it rounds if they are already strings)... but I didn't test.

Looks like it works...
```
perl -e '$a="1.23432"; $b=sprintf("%.1f", $a); print($b)'
1.2
```

starcheck/src/lib/Ska/Starcheck/Obsid.pm
taldcroft: As long as I'm trying out perl... I feel like the `substr` is not helping out here. In the unlikely event that P2 is > 10 or negative, the `substr` would give a bad result.
```
perl -e '$a="10.23432"; $b=substr($a, 0, 3); print("$b\n")'
10.
```

jeanconn:
I think your example is not super reassuring, but agree that sprintf looks to round correctly even if the thingy is already a string (I suppose that makes sense it needs to cast to print the float with decimal places anyway).
```
ska3-fido$ perl -e '$a="1.29432"; $b=sprintf("%.1f", $a); print("$b\n");'
1.3
```
Good perl reminder.

jeanconn:
Wrt "And speaking of which, is this a good time to include the guide count in the starcheck per-obsid report instead of just in the summary up top?." I went ahead and did that in some commits in this PR after Javier's.  I also just added the guide count to the per-obsid stuff at the bottom.  It turns out that chunk is not parsed by the parser anyway, so seems not a problem/complication.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -2619,6 +2619,19 @@ sub quat2radecroll {
...
+###################################################################################
+    my $self = shift;
+    my $guide_count = $self->count_guide_stars();
+
+    if ($guide_count < 4.0){
+        push @{$self->{warn}}, "Guide count of $guide_count < 4.0.";
+    }
+
+    # Also save the guide count in the figure_of_merit
+    $self->{figure_of_merit}->{guide_count} = $guide_count;
```
jeanconn, line 2634: I'm not thrilled by coding with "side-effects" like this code that also saves the guide count into the data structure.  But it does seem pretty consistent with other pieces of starcheck.

jeanconn:
I put updated output at https://icxc.cfa.harvard.edu/aspect/tmp/starcheck-pr378-v2/starcheck.html and https://icxc.cfa.harvard.edu/aspect/tmp/starcheck-pr378-v2/diff.html though I note we run into the little issue that as it gets further from NOV1521, we get more critical warnings on magnitude mismatches as the supplement has updated.

There might be value in adding an option to starcheck to use the mags from previous starcheck outputs. (just disabling the supplement seems not the right thing and keeping the critical warning still seems safe).

taldcroft:
What about outputting the ER guide star count which has been the driver here?

taldcroft:
About `>> CRITICAL: [4] Guide sum mag diff from agasc mag   0.01170`, this is definitely not critical in the way that it was when the check was first included. I think it can be demoted to caution or info.

jeanconn:
Good point about the ER bright star count.  I was missing that tree in this forest.  So, do we want that in the two places, the summary at top and the standard matter at the bottom of an obsid (though this one would only be for  ERs or we could include the ER bright star count on ERs and the OR guide count for ORs).

jeanconn:
I don't feel strongly about demoting the guide mag diff critical, just because it shouldn't come up in flight products.  For test work, demoting the warning doesn't fix the problem.

taldcroft:
I don't think we need to put the ER count in the summary, but I do think we should have both the ER count and the regular count in the detail text for ERs since both are checked (IIRC?).

taldcroft:
> I don't feel strongly about demoting the guide mag diff critical, just because it shouldn't come up in flight products. For test work, demoting the warning doesn't fix the problem.

Well at least it would let us pass test products without a Note related to this issue.

jeanconn:
If the note were based on the diff of all warnings... demoting wouldn't really help... but if it is your preference that's fine.

jeanconn:
OK.  I added the count of bright guide stars to the ER standard reporting.

taldcroft:
Can you make the latest output available somewhere? The top description should be updated with relevant permalinks for the functional testing / evaluation.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
taldcroft: I'm confused about this new checking method since I thought the scope here was basically just formatting and updated output.

Put in another way, isn't there already code that is checking for guide count < 4.0?

taldcroft (commented):
Looks basically OK apart from the one comment, but I'd like to see the output and diff.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: So for that, just a question about if we want to call the guide count calculation method twice.

starcheck/src/starcheck.pl
```diff
@@ -674,7 +675,7 @@ sub json_obsids{
     $out .= sprintf "<A HREF=\"#obsid$obs{$obsid}->{obsid}\">OBSID = %5s</A>", $obs{$obsid}->{obsid};
     $out .= sprintf " at $obs{$obsid}->{date}   ";
 
-    my $guide_count = $obs{$obsid}->count_guide_stars();
+    my $guide_count = $obs{$obsid}->{figure_of_merit}->{guide_count};
```
jeanconn, line 678: Here, by setting guide count from the value stored in the structure we save a call for the place where we were previously checking guide count up in the summary.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
taldcroft: I'm not worried about doing the guide count calculation twice since it is not very expensive, so reducing any coupling / complexity is a good thing.

Having now looked at the code I realize what is obvious to you (and makes me remember now) that there was no check at the obsid-level for too few guide stars. This only happened in the summary code. That was really my question.

But now looking at this, doesn't the 4.0 limit need to be 6.0 for ERs?  Ala
```
        my $min_num_gui = ($obs{$obsid}->{obsid} >= 38000 ) ? 6.0 : 4.0;
```

jeanconn: I think we still want to save the value in the figure of merit hash to make it available for printing.  So, not sure if it is worth it to make a new call to the counting code or not at the point of the summary check.

jeanconn:
The output and diff through 8d9e865 are now in https://icxc.cfa.harvard.edu/aspect/tmp/starcheck-pr378-v3/ (functional testing comment at top updated).  Still not sure if you want to call count_guide_stars() again for the summary check; could go either way on that.

jeanconn:
And @javierggt I apparently can't ask for your review because it is your PR but...

taldcroft (approved):
Looks good to me.

PR #381, opened 2022-01-20
prevent matplotlib warning

javierggt:
## Description

This PR prevents the warning that appeared in sot/skare3/pull/755:
```
linestyle is redundantly defined by the 'linestyle' keyword argument and the fmt string.
```
which occurs because both fmt and linestyle/color are given.

## Testing

- [ ] Passes unit tests on MacOS, linux, Windows (at least one required)
- [x] Functional testing

Fixes #

### DIFF
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -519,7 +519,7 @@ def plot_two(fig_id, x, y, x2, y2,
     fig = plt.figure(fig_id, figsize=figsize)
     fig.clf()
     ax = fig.add_subplot(1, 1, 1)
-    ax.plot_date(xt, y, fmt='-', linestyle=linestyle, color=color)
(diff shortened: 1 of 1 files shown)


jeanconn:
This one isn't *quite* ready, if I've tested correctly:

Flight:
flight_ccd_temperature

Test:
test_ccd_temperature

I think it should still explicitly define no marker in the new call.

taldcroft:
Agreed with @jeanconn here. The default for `plot_date` has `fmt='o'`.  https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.plot_date.html

javierggt:
Yup. Fixed it.

jeanconn:
Looks good to me now:

test_ccd_temperature2

This output can count as functional testing.

jeanconn (approved):

PR #382, opened 2022-01-31
Set linestyle via fmt in plot_two

jeanconn:
## Description

Set linestyle via fmt in plot_two

The previous fix to set the marker explicitly with 'marker' still had matplotlib throwing a warning because `fmt` is set and has default of 'o' in plot_date.

## Testing

- [ n/a] Passes unit tests on MacOS, linux, Windows (at least one required)
- [x] Functional testing

This correctly makes the plot, still without the markers and with the linestyles... and doesn't throw any warnings in ska3-next or current ska3 flight.

plottest3

### DIFF
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -508,7 +508,6 @@ def write_obstemps(output_dev, obstemps):
 
 
 def plot_two(fig_id, x, y, x2, y2,
-             linestyle='-', linestyle2='-',
              color='blue', color2='magenta',
              ylim=None, ylim2=None,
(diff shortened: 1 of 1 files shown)


jeanconn:
We could also just remove the linestyle options in plot_two and set fmt='-' in both places.  The calc_ccd_temps version of plot_two is not intended to be a public method.

javierggt (approved):
Simple enough. Anyway I checked it works.

taldcroft (approved):
I did not run myself but the testing description looks good.

taldcroft:
I also searched the `sot` org for external calls to `plot_two` and found none (since the API is changing).

PR #384, opened 2022-06-02
WIP: Remove 'nonstandard dither' warning on normal mon dither

jeanconn:
## Description

Remove 'nonstandard dither' warning on normal mon dither

## Interface impacts
None

### Unit tests

- [ ] No unit tests
- [ ] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests
Ran on MAY3022B schedule and 24483 no longer has nonstandard dither CAUTION and has new INFO statement as expected (mag diff warning expected for schedule after agasc update):
```
>> CAUTION : [ 4] Search spoiler.  529278336: Y,Z,Radial,Mag seps:  41  40  57 -0.8
>> CAUTION : [10] Search spoiler.  529799376: Y,Z,Radial,Mag seps: 164  55 173 -0.4
>> CAUTION : [4] Guide sum mag diff from agasc mag   0.01140
>> INFO    : Has Monitor window dither transition
>> INFO    : Monitor window special commanding meets requirements
>> INFO    : Reviewed with ACQ dither Y=0.0 Z=0.0 
```

### DIFF
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -857,6 +857,9 @@ sub check_dither {
         }
     }
 
+    my $c = find_command($self, "MP_STARCAT");
+    my @mon_stars = grep { $c->{"TYPE$_"} eq 'MON' } (1..16);
+
 
     $self->{dither_acq} = $acq_dither;
     $self->{dither_guide} = $guide_dither;
@@ -865,8 +868,16 @@ sub check_dither {
 
     # Check for standard dither
     if ($guide_dither->{state} eq 'ENAB'){
-        if ((not standard_dither($guide_dither)) or not standard_dither($acq_dither)){
-            push @{$self->{yellow_warn}}, "Non-standard dither\n";
+	if ((not standard_dither($guide_dither)) or not standard_dither($acq_dither)){
+	    # If the acquisition dither is zero and there is a mon window that's normal.
+	    if ((scalar(@mon_stars) > 0) and standard_dither($guide_dither)
+		and ($acq_dither->{ampl_p} == 0) and ($acq_dither->{ampl_y} == 0)){
+		push @{$self->{fyi}}, "Has Monitor window dither transition\n";
+	    }
+	    else{
+		push @{$self->{yellow_warn}}, "Non-standard dither\n";
+	    }
+
         }
         if (($guide_dither->{ampl_p} != $acq_dither->{ampl_p})
                 or ($guide_dither->{ampl_y} != $acq_dither->{ampl_y})){


starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -865,8 +868,16 @@ sub check_dither {
 
     # Check for standard dither
     if ($guide_dither->{state} eq 'ENAB'){
-        if ((not standard_dither($guide_dither)) or not standard_dither($acq_dither)){
-            push @{$self->{yellow_warn}}, "Non-standard dither\n";
+	if ((not standard_dither($guide_dither)) or not standard_dither($acq_dither)){
```
taldcroft, line 871: I think this is conflating guide dither and monitor commanding in a way that is not quite right. Since guide dither is the state at +8 minutes into the observation and monitor commanding ends at +5 minutes, they are basically unrelated and so should be treated separately.

Roughly speaking I think the logic should be about:
```
if catalog has MON:
  if acq_dither amplitude is not 0.0 in either component:
    red warning (this would be a benign commanding error if dither is disabled)
  else:
    FYI message (if it were me I would not emit this, but I know you like positive confirmation)

  if not standard_dither($guide_dither):
    yellow_warn "Non-standard dither"

else:
  if ((not standard_dither($guide_dither)) or not standard_dither($acq_dither)){
            push @{$self->{yellow_warn}}, "Non-standard dither\n";

taldcroft: If an observation has non-standard guide dither, then that incorrectly blocks testing of the monitor window dither amplitude and you will always get a "Non-standard dither" warning regardless of whether the "Monitor window dither transition" is correct.

taldcroft: The current test is also relying on the historical circumstance of 0x0 dither being "non-standard" to work [1]. I've always disliked "standard" vs "non-standard" dither and this criteria should not be a gating item for entering the logic to test whether the dither is 0.0 during the monitor sequence.

[1] There is no real reason we could not have decided long ago that 0x0" dither is a standard configuration and put that in the arbitrary list.

taldcroft: > where you want to go with the 0x0 thoughts

Just that the check here for correct MON dither amplitude should not call `standard_dither()` anywhere. That happens if you follow my proposed flow.

taldcroft: One thing that just occurred to me is that the non-standard dither check is being gated by the guide dither state being `ENAB`, but if for some reason dither is disabled then that should be a caution warning as well. I.e. we want to know if an observation is being scheduled with dither disabled. This warning should be emitted even if the OR says dither should be disabled since that would most likely be a mistake.

Since maintaining dither is turning into a spacecraft safety issue we need to be extra careful now.

jeanconn: Sure.  If we want to add dither-better-be-on checks that seems fine.  Seems like they would not be as needed in sparkles, though also seems like proseco and sparkles shouldn't  let you have dynamic background bonus stars if dither disabled or too small (though that is in the weeds).

PR #389, opened 2022-08-23
Skip warnings and catalog checks if no catalog

jeanconn:
## Description
Skip warnings and catalog checks if no catalog

## Interface impacts

None.

### Unit tests

- [x] No unit tests

### Functional tests

Confirmed this removes all warnings for gyro hold observations in AUG2322A.
Also removes console warnings about trying to print undefined values (n_acq not defined for no star catalog for example).


taldcroft (approved):
Looks good!

PR #397, opened 2022-10-03
Allow using maude for thermal model source data

jeanconn:
## Description

Allow using MAUDE for thermal model source data, with automatic selection of MAUDE if eng_archive data is not available.  This is handy on my laptop as I no longer bother with syncing the archive.  Adds an option to use maude for other systems when more up-to-date telemetry is desired.

While making it work on the laptop, I also noticed that the plotting routing was always using miniagasc.h5 instead of the supplied agasc_file, so added a fix for that.

## Interface impacts

Machines without AACCCDPT in their cheta archive will fallback to using MAUDE/maude for initial conditions for the ACA thermal model.  Also adds the "-maude" option to allow users to use MAUDE/maude if preferred (but does not change default unless the archive is missing or broken).

### Unit tests

- [X] No unit tests

### Functional tests

Ran on my OSX laptop fine with AACCCDPT no longer in eng_archive.

On fido tested with and without maude option and compared the without option to flight with expected results:

 /home/jeanconn/git/starcheck/sandbox_starcheck -dir /data/mpcrit1/mplogs/2022/DEC1922/oflsa -maude -out dec1922t_maude

https://icxc.cfa.harvard.edu/aspect/test_review_outputs/starcheck-pr397/dec1922t_maude.html

/home/jeanconn/git/starcheck/sandbox_starcheck -dir /data/mpcrit1/mplogs/2022/DEC1922/oflsa -out dec1922t_cxc

https://icxc.cfa.harvard.edu/aspect/test_review_outputs/starcheck-pr397/dec1922t_cxc.html

starcheck -dir /data/mpcrit1/mplogs/2022/DEC1922/oflsa -out dec1922_flight

https://icxc.cfa.harvard.edu/aspect/test_review_outputs/starcheck-pr397/dec1922_flight.html

diff

https://icxc.cfa.harvard.edu/aspect/test_review_outputs/starcheck-pr397/diff.html

### DIFF
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -157,10 +157,12 @@ def get_ccd_temps(oflsdir, outdir='out',
         logger.info("AACCCDPT not found in cheta archive.")
         have_cxc_telem = False
 
+    use_maude = False
     if maude == 1 or not have_cxc_telem:
         fetch.data_source.set('maude allow_subset=False')
         logger.info("Setting to use maude")
         stat = None
+        use_maude = True
     else:
         stat = '5min'
 
@@ -212,8 +214,11 @@ def get_ccd_temps(oflsdir, outdir='out',
     # Get temperature telemetry for 1 day prior to min(last available telem,
     # backstop start, run_start_time) where run_start_time is for regression
     # testing.
-    tlm_end_time = min(fetch.get_time_range('aacccdpt', format='secs')[1],
-                       bs_start.secs, run_start_time.secs)
+    if use_maude:
+        tlm_end_time = min(bs_start.secs, run_start_time.secs)
+    else:
+        tlm_end_time = min(fetch.get_time_range('aacccdpt', format='secs')[1],
+                           bs_start.secs, run_start_time.secs)
     tlm = get_telem_values(tlm_end_time, ['aacccdpt'], days=1, stat=stat)
     states = get_week_states(rltt, sched_stop, bs_cmds, tlm)
 
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -123,7 +123,7 @@ def get_ccd_temps(oflsdir, outdir='out',
         have_cxc_telem = False
 
     use_maude = False
-    if maude == 1 or not have_cxc_telem:
+    if maude or not have_cxc_telem:
         fetch.data_source.set('maude allow_subset=False')
         logger.info("Setting to use maude")
         use_maude = True
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -63,15 +63,6 @@ except Exception:
     VERSION = 'dev'
 
 
-USE_MAUDE = False
-
-
-def use_maude():
-    global USE_MAUDE
-    fetch.data_source.set('maude allow_subset=False')
-    USE_MAUDE = True
-
-
 def get_options():
     import argparse
     parser = argparse.ArgumentParser(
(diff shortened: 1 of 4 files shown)


starcheck/calc_ccd_temps.py
```diff
@@ -198,9 +211,13 @@ def get_ccd_temps(oflsdir, outdir='out',
     # Get temperature telemetry for 1 day prior to min(last available telem,
     # backstop start, run_start_time) where run_start_time is for regression
     # testing.
-    tlm_end_time = min(fetch.get_time_range('aacccdpt', format='secs')[1],
-                       bs_start.secs, run_start_time.secs)
-    tlm = get_telem_values(tlm_end_time, ['aacccdpt'], days=1)
+    if use_maude:
+        tlm_end_time = min(bs_start.secs, run_start_time.secs)
```
taldcroft, line 215: This assumes that MAUDE always has some telemetry within 1.0 day of when this is being run (for the usual case of reviewing loads in the future). That can fail if we miss passes for a day. 

Since the MAUDE option is not the default this is OK, but worth noting that the end time of telemetry in MAUDE can be easily obtained with a direct `maude` query without a start/stop time:
```
import maude
dat = maude.get_msids("aacccdpt")  # returns the latest sample
maude_end_time = dat["data"][0]["times"][0]
```

starcheck/calc_ccd_temps.py
```diff
@@ -150,6 +150,19 @@ def get_ccd_temps(oflsdir, outdir='out',
...
 
+    have_cxc_telem = True
+    try:
+        fetch.get_time_range('aacccdpt')
+    except KeyError:
+        logger.info("AACCCDPT not found in cheta archive.")
+        have_cxc_telem = False
+
+    use_maude = False
+    if maude == 1 or not have_cxc_telem:
```
taldcroft, line 161: There is no way to set `maude` via the Python script version of this via `main()` and `get_options()`. I'd suggest adding that and writing `if maude or not have_cxc_telem:` here.

taldcroft (commented):
Minor comments.

starcheck/calc_ccd_temps.py
jeanconn: I know you worked on the script and main() a bit for recent testing, but I have just been thinking of calc_ccd_temps.py as only useful for this function and that the script main() was historical.  Though I suppose if you are thinking about using subprocess for python from perl it starts to become more relevant again.

taldcroft: I wasn't thinking about any of that. But it's faster to just make the change suggested by the reviewer and move on than to start a discussion.

jeanconn: That's kind of dark.... I just didn't want to make a change that I couldn't figure out how to use or test.

taldcroft: If the function is so tightly coupled to starcheck that the independent command-line interface is effectively unusable, then just cut the whole thing. But in general a command-line interface to a function should cover all the function options.

jeanconn: Agreed.  I think this is a good time to cut the cmdline interface to the function.

starcheck/calc_ccd_temps.py
jeanconn: Good catch.

jeanconn:
I updated per comments and reran the testing.  I note the testing no longer shows much difference between the telemetry sources in the temperature plot as the DEC1922 loads are well underway.

taldcroft (approved):
Looks great now. I tested this on DEC2622 and it worked as expected. In particular my cheta archive was a few days old and there was a substantial difference in the propagation time. This is great and the -maude flag will be my default going forward.

PR #402, opened 2022-12-30
Replace Inline::Python with a light Python function server

taldcroft:
## Description

Driven by problems getting `Inline::Python` to work with ska3-prime, this PR replaces all the inline Python calls with a call to a new lightweight server process that runs Python function calls. This is basically the idea that @javierggt for an API server, but it users an even lighter framework (than e.g. flask) for performance. The server gets called at least thousands of times currently, though there are probably opportunities to reduce that with some code changes.

Current testing status is that it runs to completion on the DEC1922 loads and produces output that looks about right. 

Process management for the Python function server (`starcheck/server.py`) took a little time to work out but it seems to be working, more or less. There might still be room for improvement.

The new server includes reasonable exception handling.

To do:
- [x] Proper logging not print statements
- [x] Improve performance
- [x] Get a free port from the system in Perl then pass that to the Python server
- [x] Secure communications by using a random key that is shared between client and server
- [x] Improve the `get_fid_actions` sub in parse CM. I think this might be broken for running loads in the far past.
- [x] Testing

### Unit tests

- [x] No unit tests

#### Output comparison
Ran starcheck for DEC1922A loads in this branch and in master:
```
$ ./sandbox_starcheck -dir $SKA/data/mpcrit1/mplogs/2022/DEC1922/oflsa -run_start_time=2022:352:00:00:00 -out out_{pyserver,master}
```
Then:
```
ska3) ➜  starcheck git:(no-inline-python) diff out_pyserver.txt out_master.txt
1,2c1,2
<  ------------  Starcheck 13.17.1.dev21+gd6087bd.d20230101    -----------------
<  Run on Sun Jan  1 06:45:26 EST 2023 by aldcroft from daze
---
>  ------------  Starcheck 13.17.1.dev9+g6057b15    -----------------
>  Run on Sat Dec 31 10:35:06 EST 2022 by aldcroft from daze
```

On ska3-flight HEAD linux

- [x] Confirmed that ctrl-c of the sandbox starcheck script results in all perl and python processes going away.
- [x] kill -9 of a the calling perl did not kill the starcheck.server python but it times out appropriately
- [x] Did a functional test with dither manually set to run into the load and not match kadi - completed successfully with only kadi initial state used for propagation and two warning recorded in starcheck.txt
- [x] confirmed -help works
- [x] confirmed -verbose 2 3 show different levels of outputs
- [x] confirmed -max_obsids stops processing after N obsids without other side effects

**Functional and regression tests**

Ran these in https://icxc.cfa.harvard.edu/aspect/test_review_outputs/starcheck-pr402/

```
# Long week and IR Zone holds
starcheck -dir /data/mpcrit1/mplogs/2022/DEC2622/oflsa -out dec2622a_flight
/home/jeanconn/git/starcheck_noinline/sandbox_starcheck -dir /data/mpcrit1/mplogs/2022/DEC2622/oflsa -out dec2622a_test
/proj/sot/ska/bin/diff2html dec2622a_flight.txt dec2622a_test.txt > dec2622a_diff.html

# Monitor star
starcheck -dir /data/mpcrit1/mplogs/2022/NOV2822/oflsa -out nov2822a_flight
/home/jeanconn/git/starcheck_noinline/sandbox_starcheck -dir /data/mpcrit1/mplogs/2022/NOV2822/oflsa -out nov2822a_test
/proj/sot/ska/bin/diff2html nov2822a_flight.txt nov2822a_test.txt > nov2822a_diff.html

# Maneuver only loads
starcheck -dir /data/mpcrit1/mplogs/2022/OCT2422/oflsb -out oct2422b_flight
/home/jeanconn/git/starcheck_noinline/sandbox_starcheck -dir /data/mpcrit1/mplogs/2022/OCT2422/oflsb -out oct2422b_test
/proj/sot/ska/bin/diff2html oct2422b_flight.txt oct2422b_test.txt > oct2422b_diff.html

# Replan
starcheck -dir /data/mpcrit1/mplogs/2021/MAY0521/oflsa -out may0521a_flight
/home/jeanconn/git/starcheck_noinline/sandbox_starcheck -dir /data/mpcrit1/mplogs/2021/MAY0521/oflsa -out may0521a_test
/proj/sot/ska/bin/diff2html may0521a_flight.txt may0521a_test.txt > may0521a_diff.html
```

To functionally test the change in the bad pixel check, I introduced a bad pixel near a guide star and compared outputs vs master.

https://icxc.cfa.harvard.edu/aspect/test_review_outputs/starcheck-pr402/badpixcheck/jan0923_master_badpix.html#obsid25550

https://icxc.cfa.harvard.edu/aspect/test_review_outputs/starcheck-pr402/badpixcheck/jan0923_noinline_badpix.html#obsid25550

Check is still functional.

Run of full regression test set revealed latent issue with #389 fixed in 5be93b4.  I compared this PR code to master + that same fix and the regression outputs show only two tiny, understood, and acceptable changes related to the PR update to the bad pixel test.
```
********************* 2005/JUL1105/oflsc ********************
********************* 2005/AUG2705/oflsb ********************
********************* 2005/NOV0705/oflsb ********************
********************* 2005/MAR0705/oflsb ********************
********************* 2006/MAR0606/oflsc ********************
********************* 2006/NOV1306/oflsa ********************
********************* 2006/AUG0706/oflsb ********************
********************* 2006/DEC2506/oflsc ********************
********************* 2006/MAR0606/oflsc ********************
********************* 2006/NOV2006/oflsb ********************
********************* 2007/MAR0507/oflsa ********************
********************* 2007/AUG0607/oflsa ********************
********************* 2007/AUG1407/oflsa ********************
********************* 2007/DEC1007/oflsb ********************
********************* 2007/MAR0507/oflsa ********************
********************* 2007/SEP0307/oflsa ********************
********************* 2008/JUL0708/oflsb ********************
********************* 2008/MAY0508/oflsa ********************
********************* 2008/AUG1808/oflsa ********************
********************* 2008/SEP0108/oflsb ********************
********************* 2008/SEP2908/oflsb ********************
********************* 2009/APR2009/oflsc ********************
********************* 2009/FEB1609/oflsa ********************
********************* 2009/FEB2309/oflsc ********************
********************* 2009/JUL0609/oflsb ********************
********************* 2009/JUN2209/oflsb ********************
********************* 2009/NOV3009/oflsb ********************
--- /home/jeanconn/git/starcheck_noinline/test_regress/release/2009/NOV3009/oflsb/starcheck.txt	2023-01-05 17:46:27.310431000 -0500
+++ /home/jeanconn/git/starcheck_noinline/test_regress/fido.cfa.harvard.edu_48d754734f76c6d488de6d18f96188ba4380c1e4/2009/NOV3009/oflsb/starcheck.txt	2023-01-05 23:10:38.313098000 -0500
@@ -81,7 +81,7 @@
 OBSID = 11935 at 2009:339:14:45:41.729   7.5 ACQ | 5.0 GUI | Critical:11 Caution:16
 OBSID = 11936 at 2009:339:15:53:45.926   7.6 ACQ | 5.0 GUI | Critical:10 Caution: 9
 OBSID = 11937 at 2009:339:16:35:42.174   7.6 ACQ | 5.0 GUI | Critical: 9 Caution: 8
-OBSID = 11938 at 2009:339:17:00:52.174   6.5 ACQ | 5.0 GUI | Critical:11 Caution:12
+OBSID = 11938 at 2009:339:17:00:52.174   6.5 ACQ | 5.0 GUI | Critical:10 Caution:12
 OBSID = 11939 at 2009:339:17:26:02.174   7.7 ACQ | 5.0 GUI | Critical:10 Caution: 9
 OBSID = 12037 at 2009:339:17:51:13.756   7.3 ACQ | 5.0 GUI | Critical:11 Caution:11
          ------  2009:340:03:45:00.000   OBC Load Segment Begins     CL340:0304 
@@ -1905,7 +1905,6 @@
 >> CRITICAL: [ 4] Readout Size. 6x6 Should be 8x8
 >> CRITICAL: [ 5] Readout Size. 6x6 Should be 8x8
 >> CRITICAL: [ 6] Readout Size. 6x6 Should be 8x8
->> CRITICAL: [ 7] Nearby ACA bad pixel.  row, col (-319, -299), dy, dz (5, 44) 
 >> CRITICAL: [ 7] Readout Size. 6x6 Should be 8x8
 >> CRITICAL: [ 8] Readout Size. 6x6 Should be 8x8
 >> CRITICAL: [ 9] Readout Size. 6x6 Should be 8x8
********************* 2009/OCT0509/oflsa ********************
********************* 2009/DEC2109/oflsb ********************
********************* 2010/APR1110/oflsb ********************
********************* 2010/APR1210/oflsa ********************
********************* 2010/JAN1110/oflsa ********************
********************* 2010/JUL0510/oflsb ********************
********************* 2010/OCT1110/oflsb ********************
********************* 2010/OCT2510/oflsb ********************
********************* 2011/JAN1711/oflsa ********************
********************* 2011/MAR1411/oflsa ********************
********************* 2011/APR0411/oflsa ********************
********************* 2011/DEC1211/oflsa ********************
********************* 2012/JAN3012/oflsa ********************
********************* 2013/JUL2913/oflsa ********************
********************* 2014/JAN2514/oflsa ********************
********************* 2015/JAN1215/oflsb ********************
--- /home/jeanconn/git/starcheck_noinline/test_regress/release/2015/JAN1215/oflsb/starcheck.txt	2023-01-05 18:45:24.075190000 -0500
+++ /home/jeanconn/git/starcheck_noinline/test_regress/fido.cfa.harvard.edu_48d754734f76c6d488de6d18f96188ba4380c1e4/2015/JAN1215/oflsb/starcheck.txt	2023-01-05 23:52:48.696908000 -0500
@@ -1063,7 +1063,7 @@
 [ 7]  6   185210664   BOT  8x8   0.955   9.205  10.703   1782    555  20   1  120          
 [ 8]  7   185339360   BOT  8x8   0.973   8.591  10.094   -407  -2098  20   1  120          
 
->> CRITICAL: [ 6] Nearby ACA bad pixel.  row, col (-318, -296), dy, dz (2, 25) 
+>> CRITICAL: [ 6] Nearby ACA bad pixel.  row, col (-317, -296), dy, dz (2, 25) 
 >> CAUTION : [1] Guide sum mag diff from agasc mag   0.08276
 >> CAUTION : [2] Guide sum mag diff from agasc mag   0.28450
 >> CAUTION : [3] Guide sum mag diff from agasc mag   0.02349
********************* 2015/DEC1115/oflsa ********************
********************* 2015/DEC1115/oflsb ********************
```

#### Performance
This branch runs about 50% slower. For DEC1922A this means about 90 seconds in this branch vs. 60 seconds in master.

### DIFF
--- a/starcheck/server.py
+++ b/starcheck/server.py
@@ -1,11 +1,11 @@
 import importlib
 import json
 import socketserver
+import sys
 import traceback
 
-
-PORT = 44123
 HOST = "localhost"
+KEY = None
 
 
 class MyTCPHandler(socketserver.StreamRequestHandler):
@@ -24,13 +24,18 @@ class MyTCPHandler(socketserver.StreamRequestHandler):
 
         # Decode self.data from JSON
         cmd = json.loads(data)
+
+        if cmd["key"] != KEY:
+            print(f"SERVER: bad key {cmd['key']!r}")
+            return
+
         # print(f"SERVER receive func: {cmd['func']}")
         # print(f"SERVER receive args: {cmd['args']}")
         # print(f"SERVER receive kwargs: {cmd['kwargs']}")
 
         # For security reasons, only allow functions in the public API of starcheck module
         parts = cmd["func"].split(".")
-        package = '.'.join(['starcheck'] + parts[:-1])
+        package = ".".join(["starcheck"] + parts[:-1])
         func = parts[-1]
         module = importlib.import_module(package)
         func = getattr(module, func)
@@ -52,13 +57,20 @@ class MyTCPHandler(socketserver.StreamRequestHandler):
 
 
 def main():
-    # Create the server, binding to localhost on port 9999
-    with socketserver.TCPServer((HOST, PORT), MyTCPHandler) as server:
+    global KEY
+
+    # Read a line from STDIN
+    port = int(sys.stdin.readline().strip())
+    KEY = sys.stdin.readline().strip()
+
+    print("SERVER: starting on port", port)
+
+    # Create the server, binding to localhost on supplied port
+    with socketserver.TCPServer((HOST, port), MyTCPHandler) as server:
         # Activate the server; this will keep running until you
         # interrupt the program with Ctrl-C
         server.serve_forever()
 
 
 if __name__ == "__main__":
-    print("SERVER: starting on port", PORT)
     main()
--- a/starcheck/server.py
+++ b/starcheck/server.py
@@ -15,7 +15,8 @@ class PythonServer(socketserver.TCPServer):
     timeout = 2
 
     def handle_timeout(self) -> None:
-        print("SERVER: timeout")
+        print(f"SERVER: starcheck python server timeout after {self.timeout}s idle",
+              file=sys.stderr)
         # self.shutdown()  # DOES NOT WORK, just hangs
         sys.exit(1)
 
--- a/starcheck/server.py
+++ b/starcheck/server.py
@@ -11,6 +11,15 @@ KEY = None
 func_calls = collections.Counter()
 
 
+class PythonServer(socketserver.TCPServer):
+    timeout = 2
+
+    def handle_timeout(self) -> None:
+        print("SERVER: timeout")
+        # self.shutdown()  # DOES NOT WORK, just hangs
+        sys.exit(1)
+
+
 class MyTCPHandler(socketserver.StreamRequestHandler):
     """
     The request handler class for our server.
@@ -74,10 +83,11 @@ def main():
     print("SERVER: starting on port", port)
 
     # Create the server, binding to localhost on supplied port
-    with socketserver.TCPServer((HOST, port), MyTCPHandler) as server:
+    with PythonServer((HOST, port), MyTCPHandler) as server:
         # Activate the server; this will keep running until you
         # interrupt the program with Ctrl-C
-        server.serve_forever()
+        while True:
+            server.handle_request()
 
 
 if __name__ == "__main__":
--- a/starcheck/server.py
+++ b/starcheck/server.py
@@ -1,9 +1,14 @@
 import collections
 import importlib
 import json
+import logging
 import socketserver
 import sys
 import traceback
+
+from ska_helpers.logging import basic_logger
+
+logger = basic_logger(__name__, level="INFO")
 
 HOST = "localhost"
 KEY = None
@@ -33,18 +38,18 @@ class MyTCPHandler(socketserver.StreamRequestHandler):
     def handle(self):
         # self.request is the TCP socket connected to the client
         data = self.rfile.readline()
-        # print(f"SERVER receive: {data.decode('utf-8')}")
+        logger.debug(f"SERVER receive: {data.decode('utf-8')}")
 
         # Decode self.data from JSON
         cmd = json.loads(data)
 
         if cmd["key"] != KEY:
-            print(f"SERVER: bad key {cmd['key']!r}")
+            logger.ERROR(f"SERVER: bad key {cmd['key']!r}")
             return
 
-        # print(f"SERVER receive func: {cmd['func']}")
-        # print(f"SERVER receive args: {cmd['args']}")
-        # print(f"SERVER receive kwargs: {cmd['kwargs']}")
+        logger.debug(f"SERVER receive func: {cmd['func']}")
+        logger.debug(f"SERVER receive args: {cmd['args']}")
+        logger.debug(f"SERVER receive kwargs: {cmd['kwargs']}")
 
         exc = None
 
@@ -69,20 +74,29 @@ class MyTCPHandler(socketserver.StreamRequestHandler):
                 exc = traceback.format_exc()
 
         resp = json.dumps({"result": result, "exception": exc})
-        # print(f"SERVER send: {resp}")
+        logger.debug(f"SERVER send: {resp}")
 
         self.request.sendall(resp.encode("utf-8"))
 
 
 def main():
     global KEY
-
+    
     # Read a line from STDIN
     port = int(sys.stdin.readline().strip())
     KEY = sys.stdin.readline().strip()
-
-    print("SERVER: starting on port", port)
-
+    loglevel = sys.stdin.readline().strip()
+    
+    logmap = {'0': logging.CRITICAL,
+              '1': logging.WARNING,
+              '2': logging.INFO}
+    if loglevel in logmap:
+        logger.setLevel(logmap[loglevel])
+    if int(loglevel) > 2:
+        logger.setLevel(logging.DEBUG)
+        
+    logger.info(f"SERVER: starting on port {port}")    
+    
     # Create the server, binding to localhost on supplied port
     with PythonServer((HOST, port), MyTCPHandler) as server:
         # Activate the server; this will keep running until you
--- a/starcheck/server.py
+++ b/starcheck/server.py
@@ -1,3 +1,4 @@
+import collections
 import importlib
 import json
 import socketserver
@@ -6,6 +7,8 @@ import traceback
 
 HOST = "localhost"
 KEY = None
+
+func_calls = collections.Counter()
 
 
 class MyTCPHandler(socketserver.StreamRequestHandler):
@@ -33,22 +36,27 @@ class MyTCPHandler(socketserver.StreamRequestHandler):
         # print(f"SERVER receive args: {cmd['args']}")
         # print(f"SERVER receive kwargs: {cmd['kwargs']}")
 
-        # For security reasons, only allow functions in the public API of starcheck module
-        parts = cmd["func"].split(".")
-        package = ".".join(["starcheck"] + parts[:-1])
-        func = parts[-1]
-        module = importlib.import_module(package)
-        func = getattr(module, func)
-        args = cmd["args"]
-        kwargs = cmd["kwargs"]
+        exc = None
 
-        try:
-            result = func(*args, **kwargs)
-        except Exception:
-            result = None
-            exc = traceback.format_exc()
+        # For security reasons, only allow functions in the public API of starcheck module
+        if cmd["func"] == "get_server_calls":
+            # Sort func calls by the items in the counter
+            result = dict(func_calls)
         else:
-            exc = None
+            func_calls[cmd["func"]] += 1
+            parts = cmd["func"].split(".")
+            package = ".".join(["starcheck"] + parts[:-1])
+            func = parts[-1]
+            module = importlib.import_module(package)
+            func = getattr(module, func)
+            args = cmd["args"]
+            kwargs = cmd["kwargs"]
+
+            try:
+                result = func(*args, **kwargs)
+            except Exception:
+                result = None
+                exc = traceback.format_exc()
 
         resp = json.dumps({"result": result, "exception": exc})
         # print(f"SERVER send: {resp}")
--- a/starcheck/server.py
+++ b/starcheck/server.py
@@ -1,0 +1,64 @@
+import importlib
+import json
+import socketserver
+import traceback
+
+
+PORT = 44123
+HOST = "localhost"
+
+
+class MyTCPHandler(socketserver.StreamRequestHandler):
+    """
+    The request handler class for our server.
+
+    It is instantiated once per connection to the server, and must
+    override the handle() method to implement communication to the
+    client.
+    """
+
+    def handle(self):
+        # self.request is the TCP socket connected to the client
+        data = self.rfile.readline()
+        # print(f"SERVER receive: {data.decode('utf-8')}")
+
+        # Decode self.data from JSON
+        cmd = json.loads(data)
+        # print(f"SERVER receive func: {cmd['func']}")
+        # print(f"SERVER receive args: {cmd['args']}")
+        # print(f"SERVER receive kwargs: {cmd['kwargs']}")
+
+        # For security reasons, only allow functions in the public API of starcheck module
+        parts = cmd["func"].split(".")
+        package = '.'.join(['starcheck'] + parts[:-1])
+        func = parts[-1]
+        module = importlib.import_module(package)
+        func = getattr(module, func)
+        args = cmd["args"]
+        kwargs = cmd["kwargs"]
+
+        try:
+            result = func(*args, **kwargs)
+        except Exception:
+            result = None
+            exc = traceback.format_exc()
+        else:
+            exc = None
+
+        resp = json.dumps({"result": result, "exception": exc})
+        # print(f"SERVER send: {resp}")
+
+        self.request.sendall(resp.encode("utf-8"))
+
+
+def main():
+    # Create the server, binding to localhost on port 9999
+    with socketserver.TCPServer((HOST, PORT), MyTCPHandler) as server:
+        # Activate the server; this will keep running until you
+        # interrupt the program with Ctrl-C
+        server.serve_forever()
+
+
+if __name__ == "__main__":
+    print("SERVER: starting on port", PORT)
+    main()
--- a/starcheck/server.py
+++ b/starcheck/server.py
@@ -12,7 +12,7 @@ func_calls = collections.Counter()
 
 
 class PythonServer(socketserver.TCPServer):
-    timeout = 2
+    timeout = 180
 
     def handle_timeout(self) -> None:
         print(f"SERVER: starcheck python server timeout after {self.timeout}s idle",
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -28,7 +28,6 @@ use File::Basename;
 use POSIX qw(floor);
 use English;
 use IO::All;
-use Data::Dumper qw(Dumper);
 
 use RDB;
 
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -2351,7 +2351,8 @@ sub star_image_map {
 	my $self = shift;
 	my $c;
     return unless ($c = find_command($self, 'MP_STARCAT'));
-    return unless ((defined $self->{ra}) and (defined $self->{dec}) and (defined $self->{roll}));	my $obsid = $self->{obsid};
+    return unless ((defined $self->{ra}) and (defined $self->{dec}) and (defined $self->{roll}));
+    my $obsid = $self->{obsid};
 
 	# a hash of the agasc ids we want to plot
 	my %plot_ids;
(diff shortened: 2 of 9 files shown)


jeanconn:
Great and quick idea.  It seems to me that the concerns are speed and security.  Speed seems reasonable.  For security, I like that it "only allow functions in the public API of starcheck module".  Do we also need to either limit access to the server with key or obscurity?  And/or run the Python in a sandbox or container, or as unprivileged user?

taldcroft:
No, IMO. Maybe I'm being naive, but I think an attack would need this:

- Be on the HEAD network on the same machine.
- Discover the random port for the few minutes a week that starcheck is running
- Find vulnerabilities in the exposed functions that allow doing something malicious and worthwhile to an attacker. The only functions that are exposed are those within the flight `starcheck` package.
- The only credible thing I can think of is using all the memory and making the machine crash. I don't think there is any way to run arbitrary code by calling one of those functions.
- However it is a fair point that maybe we should be make an explicit allow-list of functions just to be sure. I'm not immediately sure which symbols from `starcheck.utils` are exposed.

It also just occurred to me that the starcheck script could generate a random key and pass that to the server on creation and then require that key for commands, so that might pretty much make this lock tight.

jeanconn:
Yeah, that's what I meant by "key" as the first choice.  It wouldn't need to be SSL. And you'd add obscurity by using an available port instead of the hardcoded one (which you mention as the third box in the todo items anyway).

jeanconn:
I also wasn't sure about IPC vs TCP for this kind of app.

jeanconn:
And despite my "It wouldn't need to be SSL" comment, I don't know how much slower this would be if it used standard key security via SSL/TLS.

taldcroft:
OK, now the communication between client and server is secured by a shared random key that generated by `starcheck.pl` at the beginning. I think this addresses any credible security issues.

Also the port is now randomly selected by the OS.

taldcroft:
@jeanconn - apart from details like logging, this is ready for real review. In order to improve performance I made some code changes that you need to review carefully. 

Along the way I reformatted sections of code that I was touching since I just couldn't see the logic otherwise. It might just be worth running `perltidy` at some point.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -1445,8 +1441,8 @@ sub check_star_catalog {
         my @dr;
         if ($type =~ /GUI|BOT/){
 	    foreach my $pixel (@bad_pixels) {
-		my $dy = abs($yag-$pixel->{yag});
-		my $dz = abs($zag-$pixel->{zag});
+		my $dy = abs($pixel_row-$pixel->{row}) * 5;
```
jeanconn, line 1442: For this comparison, was the previous logic incorrect?  We convert the bad pixels to yag, zag and confirm yag zag of those pixels not within dither + 25 arcsecs of star position.  Multiplying the pixel values by 5 is not that close to the previous values.  I'm wondering if it would make sense to leave this alone for this release and then either update to match the proseco logic (work in pixel space instead) or put that bad pixel check in sparkles (to be integrated in starcheck #327 ).

starcheck/src/starcheck.pl
```diff
@@ -211,6 +219,7 @@
 
 
 # First read the Backstop file, and split into components
+print "Reading backstop file $backstop\n";
```
jeanconn, line 222: Are these just added because you wanted more logging?

starcheck/src/starcheck.pl
```diff
@@ -397,7 +422,7 @@
 
 foreach my $oflsid (keys %guidesumm){
     unless (defined $obs{$oflsid}){
-	warning("OFLS ID $oflsid in Guide Summ but not in DOT! \n");
+     #warning("OFLS ID $oflsid in Guide Summ but not in DOT! \n");
```
jeanconn, line 425: Logic change.

taldcroft: Yes revert that. The new feature to limit the number of obsids during a development cycle doesn't play well with this check and generates tons of warnings.

starcheck/src/starcheck.pl
taldcroft: It was more about profiling / timing of different steps. But yes, I never like it when starcheck just hangs for a while without any sign of activity. Early on reading the various files was taking a long time... You can cut these if you want, in the current version the period of inactivity is not *too* bad.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
taldcroft: The previous logic was correct and I believe the new logic is also correct. Note that the scale factor is being applied to a delta row/col, so the worst case change is likely < 0.05 arcsec on a check of 16 + 25 arcsec. From the perspective of actual ACA operational requirements that level of difference is inconsequential. Multiplying by the mean scale factor (around 4.996?) would reduce that even more.

My only concern is if this can cause a mismatch with sparkles. But at first glance it doesn't look like sparkles checks that, so that probably can't happen. And it sounds like you know that proseco works in pixel space, so then this change is already accomplishing a good thing, no?

jeanconn: I missed that this was a delta pixel space check in this version and was just seeing pixels * 5.  I'm blaming the lack of whitespace around an arithmetic operator.

starcheck/src/starcheck.pl
jeanconn: Seems like we should keep the check then but skip it if MAX_OBSIDS is set?

starcheck/utils.py
```diff
@@ -17,119 +29,113 @@
...
-        try:
-            return func(*args, **kwargs)
-        except Exception:
-            import traceback
-            traceback.print_exc()
-            raise
-    return wrapper
+def date2secs(val):
+    """Convert date to seconds since 1998.0"""
+    out = cxotime.date2secs(val)
```
jeanconn, line 38: I note in passing that Chandra.Time still seems to be winning on speed a bit for these very limited cases.  If we are planning to modernize it as a non-namespace package, is there any value in keeping it around in "new" code?

starcheck/src/starcheck.pl
taldcroft: Sounds good.

starcheck/utils.py
taldcroft: It's important to profile and quantify the effect on overall program execution.

The Chandra.Time version is only about 5 µs faster (16 vs 21). Which means you need about a million calls to that function in order to have an appreciable impact on a runtime that is around 60 seconds. So basically the answer is "No, we should never use it in new code unless you can demonstrate through profiling that it matters".

(One of my performance PR's does currently use DateTime, but I'll be fixing that with the `convert_time_format` function which is fast enough.)

jeanconn: I did this profiling and my conclusion was just "if we are making a change, why are we picking the slower one".  If your point is that cxotime is the supported code moving forward that's fine... I just saw this working going into chandra_time as well and was trying to figure out the reason for picking the slower code.

By my profiling I agree that there's no overall effect on program execution.

taldcroft: I picked `Chandra.Time` as a stressing test case for de-namespace because it has a compiled C++ code, not so much because I want to modernize / endorse it. But eventually *every* namespace package will get the treatment, I just didn't want to discover a problem late in the process.

starcheck/utils.py
```diff
@@ -17,119 +29,113 @@
...
-@python_from_perl
 def get_data_dir():
-    sc_data = os.path.join(os.path.dirname(starcheck.__file__), 'data')
+    sc_data = os.path.join(os.path.dirname(starcheck.__file__), "data")
     return sc_data if os.path.exists(sc_data) else ""
 
 
-@python_from_perl
-def _make_pcad_attitude_check_report(kwargs):
+def _make_pcad_attitude_check_report(**kwargs):
```
jeanconn, line 107: This is an API change (which is fine... they are just a little hard to see in the changes in here so I'm commenting now for myself).

jeanconn:
I think this is really great, but common usage for me with starcheck has been to "ctrl-c" all the time when I realize I've run with the wrong options or whatever.  I just had 8 of these python servers going.  So it seems like the python server needs to check regularly if the parent process is alive and also have some kind of reasonable timeout.

jeanconn:
(well, and also tell the user not to ctrl-c but it should be robust to that).

starcheck/src/starcheck.pl
```diff
@@ -32,35 +34,42 @@
...
+Ska::Starcheck::Python::set_port($server_port);
+Ska::Starcheck::Python::set_key($server_key);
+
+# Global process ID for fork that gets set in the parent process. This is
+# used to kill the forked process when the parent finishes.
+my $pid;
+
+# Start a server that can call functions in the starcheck package
+if ($pid = fork) {} else {
+    open(SERVER, "| python -m starcheck.server");
```
jeanconn, line 59: After fiddling with this some more... it looks to me like this code was always forking and making a new Perl process and then from that making a new Python process that was sticking around even when the Perl process was killed at the end.  I don't actually understand why the Perl $pid process is needed instead of tracking the python process id.

jeanconn, line 59:
```suggestion
# Start a server that can call functions in the starcheck package                                   
my $pid = open(SERVER, "| python -m starcheck.server");
SERVER->autoflush(1);
# Send the port and key to the server                                                               
print SERVER "$server_port\n";
print SERVER "$server_key\n";
```
?

taldcroft: The Perl $pid is the only one that it knows. See comment below about child vs. parent context.

jeanconn: I don't know what you mean?  This was all "not indented".

jeanconn: Which comment?  It certainly could know the python process and I think you've moved to my suggestion.  I had these all as commits on the side but wasn't sure if just adding them (like you told me to) was a good idea without more discussion.

jeanconn:
So I guess you didn't see my comments before starting your work.  I'd hoped to save time but.
You're current implementation matches what I had except I'd moved to using signal alarm instead of trying to figure out how to actually have handle_timeout do something.

taldcroft:
Your comments did not indicate having solved the problem.

taldcroft:
Bummer, guess I wasted my entire morning. Your comments were off track, the suggestion would not have worked (applied literally), and you didn't push any commits or say that you had something which solved the problem.

jeanconn:
My comments were asking why there was a perl fork before I removed it.

jeanconn:
Sorry about the suggestions.  I have not tried those before so I cut and paste I guess that's not how it works.

jeanconn:
I also thought ball was in my court so had just reached out trying to make sure I didn't go down a path (removing perl fork) that had different problems.

taldcroft:
Well anyway the commits I have pushed should work and finish this up. Actually I think the first commit alone is enough.

jeanconn:
Oh well.  My lesson learned is just to push the commits even if possibly controversial.  And it was a little bit of a tricky problem so the independent solutions arriving at the same place is a good sign.

taldcroft:
Update to what I said. When I start the sandbox you have processes like so:
```
 7848 ttys006    0:00.00 /bin/bash ./sandbox_starcheck -dir /Users/aldcroft/ska/data/mpcrit1/mplogs/2022/DEC1922/
 7854 ttys006    0:00.09 perl -I ./starcheck/src/lib ./starcheck/src/starcheck.pl -dir /Users/aldcroft/ska/data/m
 7855 ttys006    0:04.31 python -m starcheck.server
```
If I kill the perl (7854) then that leaves an orphaned `python -m starcheck.server`. If I will the wrapper `7848` then that cleans up the python.

jeanconn:
I also will try more github suggestion practice.  I'd highlighted the whole code block to replace but apparently that didn't take.

taldcroft:
And with the server timeout commit 11fe68c then the Python cleans up even with a kill -9 of the other perl processes. I had the timeout set short 2 seconds for testing but changed it to 15 seconds for this test.

jeanconn:
But yes, my conclusion was that if the extra perl fork was not needed by *something else* then removing it had the two advantages of letting primary process kill the python server in "nominal operations" and having everything apparently die on ctrl-c.

jeanconn:
With the handle_request / handle_timeout timeout strategy (so the timeout applies after N seconds of no requests to the server) is the 2sec timeout basically fine for production?  Or should this be something like 2 minutes?

taldcroft:
I think the timeout should be the latter, at least 2 minutes. Basically we don't want any chance it can time out for normal operation.

jeanconn:
Sounds good.  I couldn't think of anything that starcheck would do at this point that *wouldn't* need Python within a couple of seconds, but it is best to just set the timeout in minutes instead of doing a full assessment of the "time between python calls".

starcheck/src/starcheck.pl
```diff
@@ -1109,8 +1158,19 @@ sub usage
...
 
-
-
+END {
+    if (defined $pid) {
+	if ($par{verbose} gt 1){
+	    my $server_calls = call_python("get_server_calls");
+	    # print the server_calls hash sorted by value in descending order
+	    print("Python server calls:");
+	    print Dump($server_calls);
```
jeanconn, line 1167: This Dump() works fine, I just am unfamiliar with it.  Is it a perl standard library thing?  Difficult search term.

taldcroft: https://perldoc.perl.org/Data::Dumper

taldcroft: Actually now I see your confusion since the `use Data::Dumper` line was commented out. So I also have no idea where that is coming from. Double Yay Perl.

jeanconn: Right.  I can't seem to get it to work in a perldl environment, Data::Dumper mentions a Dump OO method but it doesn't seem to be the same, and this Dump you have here works without import Data::Dumper... so yes, was confused.  I'm not sure if I should leave this alone with a working unknown method or just have it use Dumper (which has slightly different formatting).

jeanconn:
For the action item of " Proper logging not print statements" it is hard to know what makes sense.  The Perl side is a mess of print statements.  The new python code could have better logging, but would we really want to use utils.config_logging for this?  One can't use it to start logging the server startup, for example, because one needs to start up the Python server to do things like configure a logger.  I'm not sure how much value it would add here.

javierggt:
I am not going into the details of this PR, just following the comments, and I wanted to say that I just used it with this weeks loads without trouble.

starcheck/server.py
```diff
@@ -69,20 +74,29 @@ def handle(self):
...
+    
     # Read a line from STDIN
     port = int(sys.stdin.readline().strip())
     KEY = sys.stdin.readline().strip()
-
-    print("SERVER: starting on port", port)
-
+    loglevel = sys.stdin.readline().strip()
+    
+    logmap = {'0': logging.CRITICAL,
```
jeanconn, line 90: With regard to logging and "not worth it initially" I just ran with including the pass-throughs for verbose logging because it was easier to test.

jeanconn:
My testing is not done, but I've updated the PR with some linux functional and regression test notes.

taldcroft:
The testing looks quite thorough, thanks!

taldcroft:
About the `decode`'s, good catch. Did you search for any outstanding `decode()` calls?

Was that code actually being run? I would assume that it would fail with an exception always if it ran.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
taldcroft: Interesting, so I am changing my expectations of variations in the local plate scale with this:
```
In [9]: y1, z1 = pixels_to_yagzag(-300, -300)
In [10]: y2, z2 = pixels_to_yagzag(-300, -310)
In [11]: (z1 - z2) / 10
Out[11]: 4.9470113374697124
```
So what matters to me really is whether starcheck can fail a catalog that proseco selected as OK. I'm not quite sure off hand. @jeanconn maybe you have been looking at the code and the threshold that proseco applies? 

At the end of the day there is a fundamental point that bad pixels are defined in pixel space and stars are located in yag/zag space, so no matter how you define the distance metric there will be changes depending on the local plate scale.

jeanconn: Right.  With regard to bad pixels defined in pixel space and stars in yag zag space, agreed.  I was just thinking that the converter mention that 'center' is often more what you have in mind and for this check that is based on "star distance to a pixel" the distance to pixel center is probably more right. 

But to your point about how consistent this is with proseco, the guide star exclusion for bad pixels is a little opaque (sorry) https://github.com/sot/proseco/blob/7a75ae329a3fde4070fbe808633db2abdcf19344/proseco/guide.py#L1330 .  The candidate star positions are converted to row/col with edge https://github.com/sot/proseco/blob/master/proseco/core.py#L1253 .  It looks to me like we're adding 4 to 5 pixels padding to define the exclusion range, so I would think this could be 1 pixel less conservative than the starcheck check for some cases.  

So worst case would be that the starcheck check would give us a red warning and we'd need to double-check this.  I was thinking this should probably go as-is and we can re-examine the tolerances in the future if they cause issues.

taldcroft: >  I was thinking this should probably go as-is and we can re-examine the tolerances in the future if they cause issues.

YES!

jeanconn (approved):
This looks good to me and I'm done with my testing.

starcheck/src/starcheck.pl
```diff
@@ -440,8 +440,7 @@
 		$obs{$oflsid}->add_guide_summ($oflsid, \%guidesumm);
     }
     else {
-	my $obsid = $obs{$oflsid}->{obsid};
```
jeanconn, line 412: I probably should have left this line (the print warning statement is likely going to also spit out an "undefined $obsid" warning.  But I'll fix it in a future PR).

jeanconn:
As noted above I just found a tiny defect in a change of mine, but I still think this is fine to go.  I just don't want to merge as some of this is my code with only (recorded) self-review.

PR #403, opened 2023-01-09
set exit code 0 at the end

javierggt:
## Description

This PR sets the exit code to `0` at the end of execution.

I can confirm that the current master version return code is not zero. This causes `test_regress.sh` to fail. Someone who is more familiar with perl should check this.

## Interface impacts
None

### Unit tests

- [x] No unit tests

### Functional tests

I checked that the program runs and returns the given value using the same command as in `test_regress.sh`:
```
starcheck -dir $SKA/data/ska_testr/test_loads/2019/MAY2019/oflsa
```

I did not check an error elsewhere to make sure the error is properly set. I would need to know what error to check for.

### DIFF
--- a/starcheck/src/starcheck.pl
+++ b/starcheck/src/starcheck.pl
@@ -1301,3 +1301,5 @@ Tom Aldcroft ( taldcroft@cfa.harvard.edu )
 
 =cut
 
+exit 0
+


starcheck/src/starcheck.pl
```diff
@@ -1223,3 +1223,5 @@ =head1 AUTHOR
 
 =cut
 
+exit 0
```
taldcroft, line 1226: I think this should be at the end of the END {} block above the docs.

jeanconn: The problem with that, I think, is that END still runs after an unexpected "die", so you get zero exit status even on a failure.

taldcroft: Ah good point. Sorry, never mind.

jeanconn: I think maybe
```
END {
   my $exit_status = $?;
   ...
   exit($exit_status);
}
```

jeanconn:
Also this code does work to output zero but it didn't start with master with the merges that are the problem.

javierggt:
> Also this code does work to output zero but it didn't start with master with the merges that are the problem.

That's right. It started at another commit. I rebased it.

javierggt:
Now looking better at the code it is obvious which statement is the one that sets the RC code (to 9):
```perl
	kill 9, $pid;                    # must it be 9 (SIGKILL)?
        my $gone_pid = waitpid $pid, 0;  # then check that it's gone
    }
```

jeanconn:
I saw that nine but I wasn't 100% sure it was setting the exit code.  Either way, my suggestion is in #404 .  It seems to work for the success case and for the simplest "die" case of "-dir" not being a products directory.

jeanconn:
I was reading up more on appropriate handling of errors around END but maybe we don't care about best practices.

jeanconn:
Superseded by #404

PR #406, opened 2023-01-20
Calculate a maneuver angle based on time (as needed) for proseco

jeanconn:
## Description

Calculate a maneuver angle based on time (as needed) for proseco. This is intended to give a reasonable maneuver angle (which is a proxy for pointing uncertainty) in the case where there is a longer period in NMM such as a NMM dwell as part of the high IR zone handling.

Supersedes #398 

Fixes #

### Unit tests

- [ ] No unit tests
- [ ] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests

No functional testing.

### DIFF
--- a/starcheck/check_ir_zone.py
+++ b/starcheck/check_ir_zone.py
@@ -1,15 +1,123 @@
+from astropy.table import Table, vstack
+from functools import lru_cache
 import numpy as np
 
-import kadi.commands
-import kadi.commands.states
+import kadi.commands as kadi_commands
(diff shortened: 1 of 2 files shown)


jeanconn:
I took a few minutes to update #398 here after the kinda big starcheck server and perltidy updates, but sounds like Tom just wants to retire this idea per comments in https://github.com/sot/ska-projects/issues/140 .

PR #411, opened 2023-06-16
Set plot ymin relative to planning not penalty limit

jeanconn:
## Description

Set ACA t_ccd model temperature plot ymin relative to planning not penalty limit.  This will be useful if the penalty limit is removed.

### Unit tests

- [x] OSX

### Functional tests

I set a local version of the ACA thermal model chandra_model to have planning limit -15 and penalty limit 0C and ran on the JUL1723A schedule.

ccd_temperature_mock_15

This is the JUL1723A output with the current thermal model limits

ccd_temperature_jul1723

As expected, this PR sets the Y minimum to the minimum of the data or the planning limit - 1.  I needed to move the planning limit *lower* to see that the plot change, as most schedules have a data range in which the y min on the plot is driven by the y min of the model data.

### DIFF
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -520,7 +520,7 @@ def make_check_plots(outdir, states, times, temps, tstart, tstop, char):
     logger.info('Making temperature check plots')
     for fig_id, msid in enumerate(('aca',)):
         temp_ymax = max(char.aca_t_ccd_planning_limit, np.max(temps))
-        temp_ymin = min(char.aca_t_ccd_penalty_limit, np.min(temps))
+        temp_ymin = min(char.aca_t_ccd_planning_limit - 1, np.min(temps))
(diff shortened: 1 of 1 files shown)


starcheck/calc_ccd_temps.py
```diff
@@ -520,7 +520,7 @@ def make_check_plots(outdir, states, times, temps, tstart, tstop, char):
     logger.info('Making temperature check plots')
     for fig_id, msid in enumerate(('aca',)):
         temp_ymax = max(char.aca_t_ccd_planning_limit, np.max(temps))
-        temp_ymin = min(char.aca_t_ccd_penalty_limit, np.min(temps))
+        temp_ymin = min(char.aca_t_ccd_planning_limit - 1, np.min(temps))
```
jeanconn, line 523: We could also just auto scale the plot minimum to the data, but I figured there would be no harm in explicitly setting the default min to max range as 1 degree .

PR #412, opened 2023-06-16
Use dyn_bgd_n_faint = 2 

jeanconn:
## Description

Update starcheck to apply dyn_bgd_n_faint bonus = 2 to guide_count and imposter checking.

## Interface impacts
None.

## Testing
Tested with ska3-matlab-2023.4rc6 on fido.  This PR requires sparkles 4.22.1 .

### Unit tests

- [x] Unit test added to this as #418 

### Functional tests

Ran a set of weeks and tested the cmdline option with output from this script at
https://icxc.cfa.harvard.edu/aspect/test_review_outputs/starcheck-pr412

```
jeanconn-fido> more run.sh
# Long week and IR Zone holds
/home/jeanconn/git/starcheck/sandbox_starcheck -dir /data/mpcrit1/mplogs/2022/DEC2622/oflsa -out dec2622a_test
/proj/sot/ska3/flight/bin/starcheck -dir /data/mpcrit1/mplogs/2022/DEC2622/oflsa -out dec2622a_flight
/proj/sot/ska/bin/diff2html dec2622a_flight.txt dec2622a_test.txt > dec2622a_diff.html

# Monitor star
/home/jeanconn/git/starcheck/sandbox_starcheck -dir /data/mpcrit1/mplogs/2022/NOV2822/oflsa -out nov2822a_test
/proj/sot/ska3/flight/bin/starcheck -dir /data/mpcrit1/mplogs/2022/NOV2822/oflsa -out nov2822a_flight
/proj/sot/ska/bin/diff2html nov2822a_flight.txt nov2822a_test.txt > nov2822a_diff.html

# Maneuver only loads
/home/jeanconn/git/starcheck/sandbox_starcheck -dir /data/mpcrit1/mplogs/2022/OCT2422/oflsb -out oct2422b_test
/proj/sot/ska3/flight/bin/starcheck -dir /data/mpcrit1/mplogs/2022/OCT2422/oflsb -out oct2422b_flight
/proj/sot/ska/bin/diff2html oct2422b_flight.txt oct2422b_test.txt > oct2422b_diff.html

# Replan
/home/jeanconn/git/starcheck/sandbox_starcheck -dir /data/mpcrit1/mplogs/2021/MAY0521/oflsa -out may0521a_test
/proj/sot/ska3/flight/bin/starcheck -dir /data/mpcrit1/mplogs/2021/MAY0521/oflsa -out may0521a_flight
/proj/sot/ska/bin/diff2html may0521a_flight.txt may0521a_test.txt > may0521a_diff.html

# JUN1923 week Needs dynamic bonus
/home/jeanconn/git/starcheck/sandbox_starcheck -dir /data/mpcrit1/mplogs/2023/JUN1923/oflsa -out jun1923a_test
/proj/sot/ska3/flight/bin/starcheck -dir /data/mpcrit1/mplogs/2023/JUN1923/oflsa -out jun1923a_flight
/proj/sot/ska/bin/diff2html jun1923a_flight.txt jun1923a_test.txt > jun1923a_diff.html

# JUN0523 week Needs dynamic bonus
/home/jeanconn/git/starcheck/sandbox_starcheck -dir /data/mpcrit1/mplogs/2023/JUN0523/oflsa -out jun0523a_test
/proj/sot/ska3/flight/bin/starcheck -dir /data/mpcrit1/mplogs/2023/JUN0523/oflsa -out jun0523a_flight
/proj/sot/ska/bin/diff2html jun0523a_flight.txt jun0523a_test.txt > jun0523a_diff.html
```

### DIFF
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -325,12 +325,6 @@ sub find_command {
     return undef;
 }
 
-sub set_dyn_bgd_n_faint {
-    my $self = shift;
-    my $dyn_bgd_n_faint = shift;
-    $self->{dyn_bgd_n_faint} = $dyn_bgd_n_faint;
-}
-
 ##################################################################################
 sub set_maneuver {
     #
@@ -985,12 +979,11 @@ sub check_bright_perigee {
         }
     }
 
-    # Pass 1 to _guide_count as third arg to use the count_9th mode
+    # Pass 1 to _guide_count as fourth arg to use the count_9th mode
     my $bright_count = sprintf(
         "%.1f",
         call_python(
-            "utils._guide_count",
-            [ \@mags, $self->{ccd_temp}, 1, $self->{dyn_bgd_n_faint} ]
+            "utils._guide_count", [ \@mags, $self->{ccd_temp}, $self->{date}, 1 ]
         )
     );
     if ($bright_count < $min_n_stars) {
@@ -1219,7 +1212,6 @@ sub check_star_catalog {
                 $self->{date},
                 $dither_guide_y,
                 $dither_guide_z,
-                $self->{dyn_bgd_n_faint},
             ]
         );
     };
@@ -2840,8 +2832,7 @@ sub count_guide_stars {
     return sprintf(
         "%.1f",
         call_python(
-            "utils._guide_count",
-            [ \@mags, $self->{ccd_temp}, 0, $self->{dyn_bgd_n_faint} ]
+            "utils._guide_count", [ \@mags, $self->{ccd_temp}, $self->{date}, 0 ]
         )
     );
 }
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -979,11 +979,11 @@ sub check_bright_perigee {
         }
     }
 
-    # Pass 1 to _guide_count as fourth arg to use the count_9th mode
+    # Pass 1 to guide_count as count_9th kwarg to use the count_9th mode
     my $bright_count = sprintf(
         "%.1f",
         call_python(
-            "utils._guide_count",
+            "utils.guide_count",
             [],
             {
                 mags => \@mags,
@@ -2839,7 +2839,7 @@ sub count_guide_stars {
     return sprintf(
         "%.1f",
         call_python(
-            "utils._guide_count",
+            "utils.guide_count",
             [],
             {
                 mags => \@mags,
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -983,7 +983,14 @@ sub check_bright_perigee {
     my $bright_count = sprintf(
         "%.1f",
         call_python(
-            "utils._guide_count", [ \@mags, $self->{ccd_temp}, $self->{date}, 1 ]
+            "utils._guide_count",
+            [],
+            {
+                mags => \@mags,
+                t_ccd => $self->{ccd_temp},
+                count_9th => 1,
+                date => $self->{date},
+            }
         )
     );
     if ($bright_count < $min_n_stars) {
@@ -2832,7 +2839,14 @@ sub count_guide_stars {
     return sprintf(
         "%.1f",
         call_python(
-            "utils._guide_count", [ \@mags, $self->{ccd_temp}, $self->{date}, 0 ]
+            "utils._guide_count",
+            [],
+            {
+                mags => \@mags,
+                t_ccd => $self->{ccd_temp},
+                count_9th => 0,
+                date => $self->{date},
+            }
         )
     );
 }
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -1218,7 +1218,8 @@ sub check_star_catalog {
                 $self->{ccd_temp},
                 $self->{date},
                 $dither_guide_y,
-                $dither_guide_z
+                $dither_guide_z,
+                $self->{dyn_bgd_n_faint},
             ]
         );
     };
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -325,6 +325,12 @@ sub find_command {
     return undef;
 }
 
+sub set_dyn_bgd_n_faint {
+    my $self = shift;
+    my $dyn_bgd_n_faint = shift;
+    $self->{dyn_bgd_n_faint} = $dyn_bgd_n_faint;
+}
+
 ##################################################################################
 sub set_maneuver {
     #
@@ -980,8 +986,13 @@ sub check_bright_perigee {
     }
 
     # Pass 1 to _guide_count as third arg to use the count_9th mode
-    my $bright_count = sprintf("%.1f",
-        call_python("utils._guide_count", [ \@mags, $self->{ccd_temp}, 1 ]));
+    my $bright_count = sprintf(
+        "%.1f",
+        call_python(
+            "utils._guide_count",
+            [ \@mags, $self->{ccd_temp}, 1, $self->{dyn_bgd_n_faint} ]
+        )
+    );
     if ($bright_count < $min_n_stars) {
         push @{ $self->{warn} }, "$bright_count star(s) brighter than scaled 9th mag. "
           . "Perigee requires at least $min_n_stars\n";
@@ -2825,8 +2836,13 @@ sub count_guide_stars {
             push @mags, $mag;
         }
     }
-    return
-      sprintf("%.1f", call_python("utils._guide_count", [ \@mags, $self->{ccd_temp} ]));
+    return sprintf(
+        "%.1f",
+        call_python(
+            "utils._guide_count",
+            [ \@mags, $self->{ccd_temp}, 0, $self->{dyn_bgd_n_faint} ]
+        )
+    );
 }
 
 ###################################################################################
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -1493,7 +1493,7 @@ sub check_star_catalog {
                 my $guide_mag_warn = sprintf "[%2d] Magnitude. Guide star %6.3f\n", $i,
                   $mag;
                 if (($mag > 10.3) or ($mag < 5.2)) {
-                    push @warn, $guide_mag_warn;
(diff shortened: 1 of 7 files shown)


jeanconn:
@taldcroft I was thinking about updating this WIP to set dyn_bgd_n_faint to 2 by default but include a starcheck cmdline option (mostly if we needed to run with dyn_bgd_n_faint = 0 for transition weeks or regression).  

What are your thoughts?  I think we don't need to take the dyn_bgd_n_faint value from the pickle, though if we did we'd then want to check it against 2 anyway and make sure that dyn_bgd_n_faint wasn't set to an unexpected value of != 2.

What were you thinking on this topic?

jeanconn:
I updated the title to show that this just applies to starcheck guide_count.  I think it might make sense to merge this and then do another PR to include the bonus in the imposter calculations.

taldcroft:
This is a good example of where our infrastructure does not handle changing configurations very well. We need some data source to tell starcheck what is the expected value of `dyn_bgd_n_faint` and `dyn_bgd_dt_ccd` at a particular time. `chandra_models` is an obvious candidate.

Overall I don't think that `dyn_bgd_n_faint` should be singled out for a new command line option over the many other review criteria that are hardwired. The current situation is that for obsids after the patch uplink the review should be done using 2, -4.0 respectively since those are the flight-approved values of those constants (approved by SSAWG) that reflect the spacecraft capabilities. For the time being I'd prefer setting those constants based on the observation date being before or after the patch uplink date.

I agree with the concept of exporting values like `dyn_bgd_n_faint`, but not to the top of the starcheck output. It should go to a machine-readable JSON along with all of the other header info. That's a different PR, so for now I'd just remove that new line of starcheck header. (Again, I don't see `dyn_bgd_n_faint` being conceptually different from the T_ccd limit or any other review critieria. It is just the parameter du jour).

jeanconn:
I just figured a cmdline option was the easiest way to let me see "what would this be without the bonus" which I thought could have value for load review to me and testing during the transition.

jeanconn:
Updated to use local date-based selection and re-running tests.

taldcroft:
> "what would this be without the bonus"

I'm not sure why we need to answer that question going forward. The program has agreed that mission loads shall be planned using the bonus when dyn bgd is enabled and it is extremely unlikely that we'll ever plan observations using the legacy background.

In the event that we do need to answer that question, it can be done easily enough with a local dev version with changed constants.

starcheck/utils.py
```diff
@@ -245,9 +248,20 @@ def _yagzag_to_pixels(yag, zag):
...
+        dyn_bgd_n_faint = 0
+    else:
+        dyn_bgd_n_faint = 2
     eff_t_ccd = get_effective_t_ccd(t_ccd)
-    return float(guide_count(np.array(mags), eff_t_ccd, count_9th))
+    t_ccds_bonus = get_t_ccds_bonus(mags, eff_t_ccd, dyn_bgd_n_faint, dyn_bgd_dt_ccd)
+    return t_ccds_bonus
+
+
+def _guide_count(mags, t_ccd, date, count_9th=False):
```
taldcroft, line 262: It's never a good idea to change the order of positional arguments. I guess this is a private function but still.

starcheck/utils.py
```diff
@@ -245,9 +248,20 @@ def _yagzag_to_pixels(yag, zag):
     return row.tolist(), col.tolist()
 
 
-def _guide_count(mags, t_ccd, count_9th=False):
+def _t_ccds_bonus(mags, t_ccd, date):
+    dyn_bgd_dt_ccd = -4.0
+    if CxoTime(date).date < '2023:139':
+        dyn_bgd_n_faint = 0
+    else:
+        dyn_bgd_n_faint = 2
```
taldcroft, line 256: Perfect use case for Python ternary.

starcheck/utils.py
```diff
@@ -245,9 +248,20 @@ def _yagzag_to_pixels(yag, zag):
     return row.tolist(), col.tolist()
 
 
-def _guide_count(mags, t_ccd, count_9th=False):
+def _t_ccds_bonus(mags, t_ccd, date):
+    dyn_bgd_dt_ccd = -4.0
+    if CxoTime(date).date < '2023:139':
```
taldcroft, line 253: Add a comment about this magic date.

starcheck/utils.py
```diff
@@ -282,9 +296,10 @@ def check_hot_pix(idxs, yags, zags, mags, types, t_ccd, date, dither_y, dither_z
              star or fid info to make a warning.
     """
 
-    eff_t_ccd = get_effective_t_ccd(t_ccd)
-
-    dark = aca_dark.get_dark_cal_image(date=date, t_ccd_ref=eff_t_ccd, aca_image=True)
+    dark_props = aca_dark.get_dark_cal_props(date=date, include_image=True,
+                                             aca_image=True)
+    dark = dark_props['image']
+    dark_t_ccd = dark_props['t_ccd']
```
taldcroft, line 302: OK, looks like you are doing the next part of this project now.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -979,9 +979,13 @@ sub check_bright_perigee {
         }
     }
 
-    # Pass 1 to _guide_count as third arg to use the count_9th mode
-    my $bright_count = sprintf("%.1f",
-        call_python("utils._guide_count", [ \@mags, $self->{ccd_temp}, 1 ]));
+    # Pass 1 to _guide_count as fourth arg to use the count_9th mode
+    my $bright_count = sprintf(
+        "%.1f",
+        call_python(
```
taldcroft, line 985: `call_python` allows use of kwargs, so both `date` and `count_9th` would be good as kwargs. Otherwise just looking at this code it is not obvious what the `1` is indicating.

starcheck/utils.py
jeanconn: Well, since it was two weeks later and it wasn't going to get merged as-is anyway.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: Shouldn't they all just be kwargs?  I confess I was confused by the args vs kwargs choices with the call_python implementation.

jeanconn:
And while I'd rerun the testing with the latest updates, I'd forgotten to update the description to indicate that the cmdline option is gone and the starcheck text output is gone.  Done.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
taldcroft: Looks good now.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -979,9 +979,20 @@ sub check_bright_perigee {
         }
     }
 
-    # Pass 1 to _guide_count as third arg to use the count_9th mode
-    my $bright_count = sprintf("%.1f",
-        call_python("utils._guide_count", [ \@mags, $self->{ccd_temp}, 1 ]));
+    # Pass 1 to _guide_count as fourth arg to use the count_9th mode
```
taldcroft, line 982: This comment is no longer accurate but can be removed now with the kwargs explicitly called out.

starcheck/utils.py
```diff
@@ -13,6 +13,8 @@
 from astropy.table import Table
 from Chandra.Time import DateTime
 from chandra_aca.star_probs import guide_count, mag_for_p_acq
+from chandra_aca.dark_model import dark_temp_scale
+from sparkles import get_t_ccds_bonus
```
taldcroft, line 17: As a continuing evolution of best practices (and something @javierggt already does a lot)... In a case like this where the sparkles `get_t_ccds_bonus` function is a bit ambigous with the local `_t_ccds_bonus`, the preferred use would be:
```
import sparkles
...
blah = sparkles.get_t_ccds_bonus(...)
```

starcheck/utils.py
```diff
@@ -245,9 +248,18 @@ def _yagzag_to_pixels(yag, zag):
     return row.tolist(), col.tolist()
 
 
-def _guide_count(mags, t_ccd, count_9th=False):
+def _t_ccds_bonus(mags, t_ccd, date):
```
taldcroft, line 251: I'm not sure why `_t_ccds_bonus` and `_guide_count` are being named as private methods. They are effectively used as public elsewhere in the code.

starcheck/utils.py
```diff
@@ -297,12 +310,27 @@ def imposter_offset(cand_mag, imposter_mag):
...
+    eff_t_ccd = get_effective_t_ccd(t_ccd)
+
+    # Get the t_ccd bonus for n=dyn_bgd_n_faint of the guide stars.
+    guide_mags = []
+    guide_idxs = []
+    for idx, mag, ctype in zip(idxs, mags, types):
+        if ctype in ["BOT", "GUI"]:
+            guide_mags.append(mag)
+            guide_idxs.append(idx)
+    guide_t_ccds = _t_ccds_bonus(guide_mags, t_ccd, date)
```
taldcroft, line 323: It's not really a thing any more, but should the `t_ccd` here instead be `eff_t_ccd`? Otherwise the guide stars are getting an extra bonus of not having the penalty limit applied.

taldcroft (commented):
Finished review. There is only one substantive comment that might require re-running tests, but if so then you might as well address the code nitpicks.

taldcroft:
@jeanconn - looks good. Is there any way in the functional testing to validate the new code which applies the bonus temperatures to the dark cal? From what I can see the code looks reasonable but still. I see you put the `t_ccd` into a diagnostic dict, so is that exposed somewhere to view?

starcheck/utils.py
jeanconn: _t_ccds_bonus applies the penalty, so this is logically OK, but it looks like I should remove that from _t_ccds_bonus for clarity and instead pass the effective t_ccd.

jeanconn:
Good point on the functional testing.  I think I just did some print statements when I was adding the code - I will look to see if it makes sense to grab t_ccd from that diagnostic dict or just print out some stuff for this PR.

starcheck/utils.py
jeanconn: OK.  I renamed both methods.  I wasn't sure what to call either.  Hopefully guide_count() and apply_t_ccds_bonus() make some sense.

starcheck/utils.py
jeanconn: I ended up moving the effective_t_ccds back into guide_count instead of including it in apply_t_ccds_bonus.

jeanconn:
I think I still need @taldcroft to either agree this is done or decide that #418 is reasonable and should also be merged.

jeanconn:
Incremental updates to the testing outputs in the description are non-trivial across an agasc supplement update weekend, so I've  set the outputs to be remade and will rereview.

taldcroft (approved):
Looks great!

jeanconn:
👍  And I confirm I've re-reviewed the diff outputs and see only the expected changes from dynamic bonus.

PR #418, opened 2023-07-10
Add a unit test for dyn bgd hotpix imposter calc

jeanconn:
## Description
Add a unit test for dyn bgd hotpix imposter calc.  Maybe it is a bridge too far for the #412 PR, but we could start to accumulate Python unit tests here as we go.  This is not a particularly clean unit test; I suppose it would be better to ask sparkles what we want for dyn_bgd_n_faint and dyn_bgd_dt_ccd or some such?  But this does work as a functional test for the imposter calc.

This addresses the issue at the end of #412 that the functional testing of the imposter calculation was not demonstrated.

## Testing

I tested this with install testing into my ska3-matlab-2023.4rc6 environment.

### Unit tests

- [x] Mac

- [ ] [PLATFORM]:

### Functional tests

No functional testing.

### DIFF
--- a/starcheck/tests/test_utils.py
+++ b/starcheck/tests/test_utils.py
@@ -46,12 +46,6 @@ def test_check_dynamic_hot_pix():
     # These stars are in mag-sorted order so the bonus should be applied to the last two
     # Stars with idx 7 and 8 should have bonus-applied t_ccd
     dyn_bgd_dt_ccd = 4.0
-    for imposter in imposters2:
-        assert imposter["status"] == 0
-        if imposter["idx"] < 7:
-            assert imposter["t_ccd"] == t_ccd
-        else:
-            assert imposter["t_ccd"] == t_ccd - dyn_bgd_dt_ccd
 
     # The imposters should be the same except for t_ccd, offset, mag
     # as these dates were selected to have matching dark cal files
--- a/starcheck/tests/test_utils.py
+++ b/starcheck/tests/test_utils.py
@@ -1,0 +1,70 @@
+from starcheck.utils import check_hot_pix
+
+
+def test_check_dynamic_hot_pix():
+    # Parameters of obsid 25274 from JUL0323A
+    idxs = [1, 2, 3, 4, 5, 6, 7, 8]
+    yags = [
+        -773.135672595873,
+        2140.37683262341,
+        -1826.2356726102,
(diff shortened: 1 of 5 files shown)


jeanconn:
@taldcroft I don't know if we actually want to merge this or something like it, but it seemed a reasonable way to document what I did for functional test.

starcheck/tests/test_utils.py
```diff
@@ -0,0 +1,70 @@
...
+    date = "2023:138"
+
+    imposters1 = check_hot_pix(
+        idxs, yags, zags, mags, types, t_ccd, date, dither_y, dither_z
+    )
+    # There is no bonus or penalty applied so the temperatures should all be the same as the
+    # input t_ccd
+    for imposter in imposters1:
+        assert imposter["status"] == 0
+        assert imposter["t_ccd"] == t_ccd
```
taldcroft, line 43: Move this into the main checking loop.
```suggestion
```

taldcroft (changes requested):
This looks great! There was only one real issue with using `>=` or `<=` instead of splitting to `>` and `==`. Then I saw an opportunity to simplify the code and started typing.

jeanconn:
OK.  Thanks.  I just committed your changes and re-ran the test successfully.

jeanconn:
This is a PR _against_ #412, so my preference would be to get #412 approved without this, get this approved, and then merge them both.

taldcroft (approved):
Looks good now. I ran the test and got a pass. Not required, but probably worth cleaning up these warnings by preceding the regex string with `r`.
```
============================================= warnings summary ==============================================
starcheck/pcad_att_check.py:51
  /Users/aldcroft/git/starcheck/starcheck/pcad_att_check.py:51: DeprecationWarning: invalid escape sequence '\d'
    match = re.match('^(\d+\.\d+)\s+\|\s+(\S+)\s*$',

starcheck/pcad_att_check.py:67
  /Users/aldcroft/git/starcheck/starcheck/pcad_att_check.py:67: DeprecationWarning: invalid escape sequence '\d'
    match = re.match('^(\d+\.\d+)\s+\|\s+(\S+)\s+(\S+)\s+(\S+)\s+(\S+)\s*', line)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

PR #408, opened 2023-05-31
Calculate maneuver times and angle from cmds, replace maneuver summary file

jeanconn:
## Description
Calculate maneuvers from backstop cmds and replace maneuver summary as source of maneuver times and angle.

This fixes the known issue that the maneuver summary file is incorrect for observations with nonzero MANSTART.
Fixes #409 

## Interface impacts

Starcheck text output changed slightly, no impact to parsing tools:
- Maneuver end time changed by 10 seconds (new value is correct)

## Testing
Tested with ska3-matlab 2023.4rc6 on fido, though for these changes no special test setup should be required.

### Unit tests

- [x] No unit tests

### Functional tests

Used the "new" regression test set we used to review https://github.com/sot/starcheck/pull/402 plus two weeks with MANSTART values that caused issues with maneuver end times in flight starcheck
```
jeanconn-fido> more run.sh
jeanconn-fido> more run.sh
# Long week and IR Zone holds
/home/jeanconn/git/starcheck/sandbox_starcheck -dir /data/mpcrit1/mplogs/2022/DEC2622/oflsa -out dec2622a_test
/proj/sot/ska3/flight/bin/starcheck -dir /data/mpcrit1/mplogs/2022/DEC2622/oflsa -out dec2622a_flight
/proj/sot/ska/bin/diff2html dec2622a_flight.txt dec2622a_test.txt > dec2622a_diff.html

# Monitor star
/home/jeanconn/git/starcheck/sandbox_starcheck -dir /data/mpcrit1/mplogs/2022/NOV2822/oflsa -out nov2822a_test
/proj/sot/ska3/flight/bin/starcheck -dir /data/mpcrit1/mplogs/2022/NOV2822/oflsa -out nov2822a_flight
/proj/sot/ska/bin/diff2html nov2822a_flight.txt nov2822a_test.txt > nov2822a_diff.html

# Maneuver only loads
/home/jeanconn/git/starcheck/sandbox_starcheck -dir /data/mpcrit1/mplogs/2022/OCT2422/oflsb -out oct2422b_test
/proj/sot/ska3/flight/bin/starcheck -dir /data/mpcrit1/mplogs/2022/OCT2422/oflsb -out oct2422b_flight
/proj/sot/ska/bin/diff2html oct2422b_flight.txt oct2422b_test.txt > oct2422b_diff.html

# Replan
[...]

PR #422, opened 2023-07-21
WIP: Set maneuver match off commanded obsid

jeanconn:
## Description
Set maneuver match off commanded obsid

This fixes the issue that after #408 , starcheck won't run on old products where the "dot obsid" or oflsid doesn't match the commanded obsid for many ERs.

### Unit tests

- [ ] No unit tests
- [ ] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests

No functional testing.


jeanconn:
I think it makes sense to move to the time matching.  The only problem with time matching that I see was basically all of the previous "fixes" that were in there to make time matching work (not needed now)

PR #429, opened 2023-08-11
WIP: Work with optional penalty limit in model spec

jeanconn:
## Description
Work with optional penalty limit in model spec

Fixes #

### Unit tests

- [ ] No unit tests
- [ ] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests

No functional testing.

### DIFF
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -535,8 +535,9 @@ def make_check_plots(outdir, states, times, temps, tstart, tstop, char):
                                figsize=(9, 5),
                                )
         ax = plots[msid]['ax']
-        plots[msid]['ax'].axhline(y=char.aca_t_ccd_penalty_limit,
-                                  linestyle='--', color='g', linewidth=2.0)
+        if char.aca_t_ccd_penalty_limit is not None:
+            plots[msid]['ax'].axhline(y=char.aca_t_ccd_penalty_limit,
+                                      linestyle='--', color='g', linewidth=2.0)
         plots[msid]['ax'].axhline(y=char.aca_t_ccd_planning_limit,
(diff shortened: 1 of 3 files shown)


starcheck/calc_ccd_temps.py
```diff
@@ -61,10 +60,28 @@
 except Exception:
     VERSION = 'dev'
 
+CM_ACA_MODEL_DATA, CM_ACA_MODEL_INFO = (
+    chandra_models.get_data('chandra_models/xija/aca/aca_spec.json'))
```
taldcroft, line 64: Best to avoid "heavy" functions that might fail in the import. I've become of fan of `lru_cache` for stuff like this. These can be written without the `get_` in front because conceptually they are more like your original global constants.
```
@functools.lru_cache()
def aca_model_and_info():
    model, info = chandra_models.get_data('chandra_models/xija/aca/aca_spec.json')
    return model, info

@functools.lru_cache()
def aca_model_spec():
    model, _ = aca_model_and_info()
    return json.loads(model)

@functools.lru_cache()
def chandra_models_version():
    _, info = aca_model_and_info()
    return info["version"]
```

taldcroft (commented):
I know this is a WIP but I couldn't help looking at it and then I had an idea.

starcheck/calc_ccd_temps.py
jeanconn: Sure.  I generally avoid "heavy" functions on import but did consider this in context (calc_ccd_temps only used by starcheck and we have no alternate strategy in mind if getting the model doesn't work).  But if the lazy imports are still considered nicer given context I'll update to that before testing.

PR #430, opened 2023-09-21
WIP: Update to use read_or_list_full

jeanconn:
## Description

Update to use read_or_list_full

Fixes #

### Unit tests

- [ ] No unit tests
- [ ] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests

No functional testing.

### DIFF
--- a/starcheck/calc_ccd_temps.py
+++ b/starcheck/calc_ccd_temps.py
@@ -37,7 +37,7 @@ import kadi.commands
 import kadi.commands.states as kadi_states
 import xija
 from chandra_aca import dark_model
-from parse_cm import read_or_list
+from parse_cm import read_or_list_full
 from chandra_aca.drift import get_aca_offsets
 from ska_helpers import chandra_models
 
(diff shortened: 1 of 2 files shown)


starcheck/calc_ccd_temps.py
```diff
@@ -234,7 +234,10 @@ def get_ccd_temps(oflsdir, outdir='out',
     make_check_plots(outdir, states, ccd_times, ccd_temps,
                      tstart=bs_start.secs, tstop=sched_stop.secs)
     intervals = get_obs_intervals(sc_obsids)
-    obsreqs = None if orlist is None else {obs['obsid']: obs for obs in read_or_list(orlist)}
```
taldcroft, line 237: Not an actual review but you can keep the ternary with 
```
obsreqs = None if orlist is None else read_or_list_full(orlist)[0]
```

jeanconn: 👍  Though with regard to only taking or caring about some of the returned tuple of a function -- which is preferrred?  Indexing into the tuple or explicitly discarding with `_`?  I'm fine with either.

taldcroft: It depends. In general prefer discarding with `_`, but in this case keeping the ternary structure using indexing works better.

taldcroft (approved):
Nice. I like the change to consistent use of `obsreqs` which is a much better name.

PR #452, opened 2025-03-10
Add checks for no dither and small dither observations

jeanconn:
## Description

Add checks for no dither and small dither observations and remove penalty-limit-based effective t_ccd calculation.

The idea is that observations with no dither should have creep-away and should be reviewed without the benefit of the dynamic background bonus applied to guide count.

This PR also adds a new warning for observations with non-zero dither less than 8 arcsec in both axes as that is unexpected at this point and would similarly not benefit from dynamic background.

A change has also been added to have the dither periods match those currently in use such that there should be no "Non-standard dither" warnings unless true non-standard patterns are used.

This PR also removes the effective ccd temperature calculation as this is no longer needed with the deprecation of the penalty limit.  This change is effectively a no-op.

### Unit tests

- [x] Mac
```
(ska3) flame:starcheck jean$ git rev-parse HEAD
ceec84615d69de374199211d55acefac3a1d50be
(ska3) flame:starcheck jean$ pytest
========================================================================== test session starts ==========================================================================
platform darwin -- Python 3.12.8, pytest-8.3.4, pluggy-1.5.0
rootdir: /Users/jean/git
configfile: pytest.ini
plugins: anyio-4.7.0, timeout-2.3.1
collected 14 items                                                                                                                                                      

starcheck/tests/test_state_checks.py .............                                                                                                                [ 92%]
starcheck/tests/test_utils.py .                                                                                                                                   [100%]

========================================================================== 14 passed in 3.98s
```
- [ ] [PLATFORM]:

### Functional tests

A set of weeks was run in regression that shows the expected changes.

https://icxc.cfa.harvard.edu/aspect/test_review_outputs/starcheck/starcheck-pr452/

These outputs also include functional review.  Week JAN0325A shows the recent no dither observations have new info statements that note the correct configurations.
```
+>> INFO    : Dither disabled or 0 - dynamic background bonus disabled for guide count.
+>> INFO    : Dither disabled or 0 - properly configured with creep-away
```

A bunch of schedules include new removal of non-standard dither warnings as these were due to some confusion in the code about what was expected for the period of the 8-arcsec half-amplitude dither in some cases.  This is actually a bug fix in this code.  The MAR0325A week includes observations with the reduced dither in one axis on ACIS-I (more rare).
```
->> CAUTION : Non-standard dither
```

The MAR1723 week shows a new warnings  intended for small dither configs and the guide count reduced due to dynamic background bonus being disabled for 4x4.
```
+>> CRITICAL: Guide count of 3.4 < 4.
+>> WARNING : Small (< 8 arcsec) dither -- double-check
+>> INFO    : Dither disabled or 0 - dynamic background bonus disabled for guide count.
```

The MAR1723 week also shows more of these warnings
```
+>> CAUTION : Non-standard dither
```
As the standard dither checks have not been set to be backward compatible with observations before the fall of 2023.

To functionally check that the dynamic background bonus is disabled for the dither disabled observations, I just added some temporary print statements in starcheck.utils to print the bonus applied temperatures, and reviewed this for the dither disabled observations.

Here is the output and t_ccds in master
```
Checking star catalog for obsid 29897
date: 2025:006:12:11:44.865
t_ccds_bonus: [-5.36332083 -5.36332083 -5.36332083 -9.36332083 -9.36332083]
mags: [7.031, 7.857, 7.968, 8.575, 8.889]
count_9th: 0
```

And the output and t_ccds in this branch.  Note just the change in the effective t_ccd for the last two stars.
```
Checking star catalog for obsid 29897
date: 2025:006:12:11:44.865
t_ccds_bonus: [-5.36332083143338, -5.36332083143338, -5.36332083143338, -5.36332083143338, -5.36332083143338]
mags: [7.031, 7.857, 7.968, 8.575, 8.889]
count_9th: 0
```

### DIFF
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -810,12 +810,12 @@ sub standard_dither {
     my %standard_dither_y = (
         20 => 1087.0,
         16 => 1414.2,
-        8 => 1000.0
+        8 => 707.1
     );
     my %standard_dither_p = (
         20 => 768.6,
         16 => 2000.0,
-        8 => 707.1
+        8 => 1000.0
     );
 
     my $ampl_p = int($dthr->{ampl_p} + 0.5);
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -2982,22 +2982,6 @@ sub set_ccd_temps {
           sprintf("CCD temperature exceeds %.1f C\n", $ccd_temp_red_limit_round);
     }
 
-    # Add info for having a penalty temperature too
-    if (defined $config{ccd_temp_yellow_limit}) {
-        if ($self->{ccd_temp} > $config{ccd_temp_yellow_limit}) {
-            push @{ $self->{fyi} },
-              sprintf("Effective guide temperature %.1f C\n",
-                $self->{ccd_temp});
-
-        }
-        if ($self->{ccd_temp_acq} > $config{ccd_temp_yellow_limit}) {
-            push @{ $self->{fyi} },
-              sprintf("Effective acq temperature %.1f C\n",
-                $self->{ccd_temp_acq});
-
-        }
-    }
-
 # Clip the acq ccd temperature to the calibrated range of the grid acq probability model
     # and add a yellow warning to let the user know this has happened.
     if (($self->{ccd_temp_acq} > 2.0) or ($self->{ccd_temp_acq} < -15.0)) {
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -2900,12 +2900,13 @@ sub check_guide_count {
 ###################################################################################
     my $self = shift;
 
-    # Disable dynamic background if guide dither is disabled or amplitude is zero
+    # Disable dynamic background if guide dither is disabled or amplitude is small
     my $guide_dither = $self->{dither_guide};
     my $dyn_bgd;
+    my $guide_dither_ampl_y = int($guide_dither->{ampl_y} + 0.5);
+    my $guide_dither_ampl_p = int($guide_dither->{ampl_p} + 0.5);
     if (   ($guide_dither->{state} eq 'DISA')
-        or (($guide_dither->{ampl_y} == 0)
-        and ($guide_dither->{ampl_p} == 0)))
+        or (($guide_dither_ampl_y == 0) and ($guide_dither_ampl_p == 0)))
     {
         $dyn_bgd = 0;
     }
@@ -2915,7 +2916,8 @@ sub check_guide_count {
 
     # If dyn_bgd is disabled, push an info statement
     if ($dyn_bgd == 0) {
-        push @{ $self->{fyi} }, "Dynamic background bonus disabled - OFF or 0 guide dither.\n";
+        push @{ $self->{fyi} },
+        "Dither disabled or 0 - dynamic background bonus disabled for guide count.\n";
     }
 
     my $guide_count = $self->count_guide_stars($dyn_bgd);
--- a/starcheck/src/lib/Ska/Starcheck/Obsid.pm
+++ b/starcheck/src/lib/Ska/Starcheck/Obsid.pm
@@ -652,10 +652,41 @@ sub check_dither {
         }
     }
 
+    # Seed the max dither amplitude with the amplitude 8 minutes into the observation
+    $guide_dither->{ampl_y_max} = $guide_dither->{ampl_y};
+    $guide_dither->{ampl_p_max} = $guide_dither->{ampl_p};
+
+    # Check for dither changes during the observation.
+    # This also updates the maximum amplitude for any observation that has dither
+    # changes.
+    # ACA-003
+    if (not defined $obs_tstop) {
+        push @{ $self->{warn} },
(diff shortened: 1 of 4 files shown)


starcheck/utils.py
```diff
@@ -300,10 +300,14 @@ def guide_count(mags, t_ccd, count_9th, date):
                   penalty, so input t_ccd should not have penalty applied)
     :param count_9th: bool or 0 or 1 for whether to use count_9th mode
     :param date: date of observation
+    :param dyn_bgd: whether to apply dynamic background bonus
     :returns: fractional guide_count as float
     """
-    eff_t_ccd = get_effective_t_ccd(t_ccd)
-    t_ccds_bonus = apply_t_ccds_bonus(mags, eff_t_ccd, date)
+    if dyn_bgd:
+        eff_t_ccd = get_effective_t_ccd(t_ccd)
```
taldcroft, line 307: Get rid of effective_t_ccd per conversation.

jeanconn:
Technically there's still some repetition rounding the dither amplitudes to the nearest int, but I think reorganizing around that could go in future PR.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -2861,7 +2883,28 @@ sub quat2radecroll {
...
     my $self = shift;
-    my $guide_count = $self->count_guide_stars();
+
+    # Disable dynamic background if guide dither is disabled or amplitude is small
+    my $guide_dither = $self->{dither_guide};
+    my $dyn_bgd;
+    my $guide_dither_ampl_y = int($guide_dither->{ampl_y} + 0.5);
+    my $guide_dither_ampl_p = int($guide_dither->{ampl_p} + 0.5);
+    if (   ($guide_dither->{state} eq 'DISA')
+        or (($guide_dither_ampl_y == 0) and ($guide_dither_ampl_p == 0)))
```
taldcroft, line 2893: I thought at some point we were going to set dyn_bgd=0 for dither < 8 in both pitch ,yaw. E.g. if dither is (4, 4) then dyn bgd is not effective. We recognize that this is stricter than MATLAB tools / sparkles, but what we want see for guide count in starcheck is the more "truthful" evaluation without the dyn bgd bonus. This saves us from doing the calculation by hand, which is what my action would be from the warning issued by the earlier code.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -686,6 +686,33 @@ sub check_dither {
         }
     }
 
+    # Small dither check for with creep-away
```
taldcroft, line 689: Move this comment to where it actually applies, probably before line 698 or 703. But also make it more useful, again emphasizing *why* not *what* in code comments. Imagine me reading this in two years, so either provide the background or at least a link to the PR which has at least some hint of why.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -2941,13 +2987,13 @@ sub set_ccd_temps {
         if ($self->{ccd_temp} > $config{ccd_temp_yellow_limit}) {
             push @{ $self->{fyi} },
               sprintf("Effective guide temperature %.1f C\n",
```
taldcroft, line 2989: I assume you are leaving the `Effective` here for API consistency?

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: We went back and forth on this.  It sounds like what I heard as the final strategy (compute guide count with dynamic background off for zero dither but give at least some kind of warning for the completely unexpected state of small dither) isn't quite what you heard - which is fine.  The middle ground of 4x4 dither should just not happen at all so also fine to update this calculation for the < 8 regime.  Sparkles is just set for the == 0 (rounded) case in https://github.com/sot/sparkles/pull/221 so that could also get changed or left alone.  Either way.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: Nope, just missed this in my search.  I'll update this and double-check the parser.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
```diff
@@ -652,10 +652,41 @@ sub check_dither {
...
     }
 
+    # Seed the max dither amplitude with the amplitude 8 minutes into the observation
+    $guide_dither->{ampl_y_max} = $guide_dither->{ampl_y};
+    $guide_dither->{ampl_p_max} = $guide_dither->{ampl_p};
+
+    # Check for dither changes during the observation.
+    # This also updates the maximum amplitude for any observation that has dither
+    # changes.
+    # ACA-003
```
jeanconn, line 662: This chunk is moved not new - I just thought the check for changes should happen earlier - though explicitly this is the only check that *uses* the maximum amplitude of dither if there are changes during the observation as well.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: Ok, I set the starcheck count to just use the < 8 check here.

starcheck/src/lib/Ska/Starcheck/Obsid.pm
jeanconn: Actually this doesn't even hit the parser because these are just in the "warnings" section.  This whole block could go so I cut it.

taldcroft (approved):
Looks good, thanks!

